PR #29, opened 2020-10-02
Avoid Hard-Coding Stop Words (used NLTK Corpus Instead)

arnavn101:
Previously, the `STOP_WORDS` in  `tokenizer.py` were hardcoded, which may not be the best approach. So instead, the NLTK corpus provides a wider array of stopwords that could provide better pre-processing results.

### DIFF
--- a/src/python/txtai/tokenizer.py
+++ b/src/python/txtai/tokenizer.py
@@ -4,16 +4,18 @@ Text tokenization methods
 
 import re
 import string
+import nltk
+from nltk.corpus import stopwords
+nltk.download('stopwords')
+
 
 class Tokenizer(object):
     """
     Text tokenization methods
     """
 
-    # English Stop Word List (Standard stop words used by Apache Lucene)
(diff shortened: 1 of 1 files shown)


davidmezzetti:
Thank you for the pull request!

The current stop word list mirrors a very limited stop word list used by Elasticsearch/Apache Lucene. I went back and forth using the NLTK stop words list and found it to be too broad for our current applications. [paperai](https://github.com/neuml/paperai) uses the current default list and it works well. codequestion uses a slightly [different tokenizer](https://github.com/neuml/codequestion/blob/master/src/python/codequestion/tokenizer.py). 

Are you running into a specific issues with the stop words? A custom tokenizer can be implemented to use NLTK if that works better for your specific use case. Above example from codequestion does that.

davidmezzetti:
Going to close this pull request, please respond or open an issue if there is further discussion on this regarding the use case.

backnotprop:
I think configuration around stop words would be useful.

davidmezzetti:
This would be a straightforward change. The vectorizer methods no longer use the tokenizer unless it's manually enabled though. But we certainly could have a pipeline config parameter for a custom stop word set in that case.

PR #32, opened 2020-10-20
Update 03_Build_an_Embeddings_index_from_a_data_source.ipynb

csheargm:
display(HTML(df.to_html(index=False))) works in a browser development environment, but will display a message like this "<IPython.core.display.HTML object>" in a local environment. Added suggestions to those who run the script in vscode or local command line environment. Instead of displaying the result to a browser, write the html result into a local file, and then use the browser to open the html file to view the result.


davidmezzetti:
Thank you for the contribution!

PR #33, opened 2020-10-20
Update 04_Extractive_QA_with_Elasticsearch.ipynb

csheargm:
In vscode or local dev environment, the original code will not start the elastic search server. By launching the elastic search server in a separate command line window, the rest of the script is able to connect the elastic search server and continues successfully.


davidmezzetti:
Thank you for the contribution!

PR #122, opened 2021-10-10
Fix incorrect import in Notebooks

hi019:
This fixes `ModuleNotFoundError: No module named 'txtai.extractor'` in the Notebooks


hi019:
No problem!

PR #287, opened 2022-05-29
the key should be writable instead of path

csnelsonchu:
Seems like a copy and paste error. Found this typo during installation


davidmezzetti:
Thank you for fixing the typo!

PR #330, opened 2022-08-26
Support for Qdrant as an embedding backend

generall:
Hi @davidmezzetti, thanks a lot for a great project!

But I believe it could be even better with proper vector search engine under the hood :wink: 
I also believe that Qdrant is the best choice for that.

Here are some of the potential benefits it could bring:

- eliminate the need for explicit index rebuilding - Qdrant manages the vector storage in a way, that no matter updates or indexation - it always available for writes and searches.
- Scale more easily, decouple storage and encoders - it usually requires very different machines to serve encoding and search, so qdrant integration will solve that.
- possibility to make txtai stateless - I have noticed, that txtai uses `path: /tmp/index` for storing index and configuration, that could be a potential problem in case of e.g. k8s deployment, volume management could become too complicated. With dedicated storage backend it could be possible to place the whole state of the application into there.

This PR is an example of how Qdrant can potentially work with txtai, there are still some rough spots, e.g. saving offsets into the configuration: `self.config["offset"] += new`.

Would love to hear your thoughts,
Thanks!

### DIFF
--- a/examples/similarity_api_with_qdrant/config.yml
+++ b/examples/similarity_api_with_qdrant/config.yml
@@ -1,0 +1,17 @@
+# Index file path
+path: /tmp/index
+
+# Allow indexing of documents
+writable: True
+
+# Embeddings index
+embeddings:
+  path: sentence-transformers/all-MiniLM-L6-v2
+  backend: qdrant
+  qdrant:
+    collection_name: Example
+    connection:
+      host: localhost
+    collection:
+      hnsw_config:
+        ef_construct: 150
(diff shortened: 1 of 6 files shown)


davidmezzetti:
Thank you for the pull request, the kind words and thinking of txtai!

Integrations with external vector/database stores is an important task on the roadmap but there are a couple paths under consideration. There will be upcoming changes to make this easier and not require core txtai changes (i.e. backends can be imported from another package). The integration will look similar but different with that approach. So merging this at this time would be premature until those details are worked out.

generall:
Good to know that txtai moves in this direction! Please consider this PR as proof of concept and demonstration of the interfaces then. Also feel free to close it.

If you would need any help from our side, just let us know;)

davidmezzetti:
Once again, appreciate it! I'll close this for now and refer back to it in the future when txtai is ready.

PR #346, opened 2022-09-23
Fix docker command "--it"

lipusz:
Hi David,

I ran into this while I was playing with txtai.

The current command generates

```
unknown flag: --it
See 'docker run --help'
```
when executing on Linux.


davidmezzetti:
Nice catch, thank you for the fix!

PR #348, opened 2022-10-01
update syntax for ES 8.x

nickchomey:
ES 8.x deprecated the `body` parameter, in favour of passing parameters directly to the query. I'm not sure if this will work retroactively with ES 7.x, but the python elasticsearch that is installed in this Example is for the most recent version (8.4.2 as of now).


davidmezzetti:
Thanks! I see the query syntax change but the code to install ES 8.x doesn't appear to be in the change. Given this is a notional example, I would just fully upgrade the notebook to the latest stable version of ES and match the query syntax appropriately vs worrying about ES 7.x.

PR #397, opened 2022-12-12
Modified logger configuration in __init__.py (libraries shouldn't modify root logger)

adin786:
Very minor tweak suggested to`__init__.py` only.

I've been getting duplicate log messages when I import `txtai` into my code.  It turns out txtai modifies Python's root logger in it's `__init__.py` using `logging.basicConfig()`.  However it is recommended that "importable" libraries do not configure log handlers/formatters, and that they should leave that to the "application" code.

So I've edited this line to better reflect the recommendations from the [official python docs](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library).

### DIFF
--- a/src/python/txtai/__init__.py
+++ b/src/python/txtai/__init__.py
@@ -4,8 +4,13 @@ Version string
 
 import logging
 
-# Set default logging format
-logging.basicConfig(format="%(asctime)s [%(levelname)s] %(funcName)s: %(message)s")
+# Configure logging for the txtai package
+# We should use logging.basicConfig() here as it modifies the root logger.  But
+# log handlers/formatters should only be tweaked in the application code.  Ref:
+#   - https://stackoverflow.com/a/27017068/19357935
+#   - https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library
+logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


davidmezzetti:
Sorry to hear you are getting duplicate log messages. Thank you for the fix!

While I appreciate the thorough explanation, I prefer to keep these types of longer comments in GH issues/commit logs. Otherwise, the code can get messy over time. 

Can you make the following minor condensation of the comment and I'll go ahead and merge this in. 

```python
# Configure logging per standard Python library recommendations
logger = logging.getLogger(__name__)
logger.addHandler(logging.NullHandler())
```

adin786:
I've committed the tweak suggested.  Thanks for the rapid response. 👍🏼 

I should've said if it turns out that the log messages are useful for tests/debugging (I imagine they are), then I guess those tests should be considered as "application" code and should be setting up their own log handlers etc where needed.  I'm still getting comfortable with logging best practises in Python but I understand that my change above brings `txtai` in line with how users expect importable packages to behave w.r.t. logging.

davidmezzetti:
All sounds good and makes sense that txtai shouldn't change logging behavior by default. Thanks again.

Issue #409, opened 2023-01-16
40_Text_to_Speech_Generation example notebook throws a ContextualVersionConflict error


hsm207:
Running this cell in the [tts example notebook ](https://colab.research.google.com/github/neuml/txtai/blob/master/examples/40_Text_to_Speech_Generation.ipynb) on google colab:

```
%%capture

from txtai.pipeline import TextToSpeech

# Create text-to-speech model
tts = TextToSpeech()
```

throws the following exception:

```
ContextualVersionConflict                 Traceback (most recent call last)
[<ipython-input-2-fd4b30fd0bec>](https://localhost:8080/#) in <module>
----> 1 from txtai.pipeline import TextToSpeech
      2 
      3 # Create text-to-speech model
      4 tts = TextToSpeech()

11 frames
[/usr/local/lib/python3.8/dist-packages/txtai/pipeline/__init__.py](https://localhost:8080/#) in <module>
     13 from .text import *
     14 from .tensors import Tensors
---> 15 from .train import *

[/usr/local/lib/python3.8/dist-packages/txtai/pipeline/train/__init__.py](https://localhost:8080/#) in <module>
      5 from .hfonnx import HFOnnx
      6 from .hftrainer import HFTrainer
----> 7 from .mlonnx import MLOnnx

[/usr/local/lib/python3.8/dist-packages/txtai/pipeline/train/mlonnx.py](https://localhost:8080/#) in <module>
      6 
      7 try:
----> 8     from onnxmltools import convert_sklearn
      9     from onnxmltools.convert.common.data_types import StringTensorType
     10 

[/usr/local/lib/python3.8/dist-packages/onnxmltools/__init__.py](https://localhost:8080/#) in <module>
     23 from .convert import convert_h2o
     24 from .convert import convert_catboost
---> 25 from .utils import load_model
     26 from .utils import save_model

[/usr/local/lib/python3.8/dist-packages/onnxmltools/utils/__init__.py](https://localhost:8080/#) in <module>
      1 # SPDX-License-Identifier: Apache-2.0
      2 
----> 3 from .main import load_model
      4 from .main import save_model
      5 from .main import set_model_version

[/usr/local/lib/python3.8/dist-packages/onnxmltools/utils/main.py](https://localhost:8080/#) in <module>
      2 
      3 import os
----> 4 from ..proto import onnx_proto
      5 from ..convert.common import utils as convert_utils
      6 from os import path

[...]

PR #410, opened 2023-01-16
Add note to restart kernel if running in Google Colab

hsm207:
Context: #409


davidmezzetti:
This happens in a lot of notebooks and is a bit unpredictable. Actually pretty common for many Colab notebooks that update a package that's already installed. 

Perhaps it's better suited somewhere in the documentation near the examples or notebook section.

davidmezzetti:
Thanks for this update. I haven't forgotten about this. It's given me the idea that the docs should have a FAQ section as there are a couple other commons things that come up.

davidmezzetti:
I just added a FAQ section to the documentation. Can you add this there?

Sorry for the bouncing around on this but having these types of issues in one place would be good for the user experience.

davidmezzetti:
Thank you for the update!

PR #437, opened 2023-02-21
add-save-load-capability-to-api

berkiney:


davidmezzetti:
Hello, thank you for the PR!

Can you please explain the use case your trying to solve? Applications automatically save to disk after each index/upsert operation.

Issue #423, opened 2023-02-08
Modify translation pipeline langdetect parameter to accept language detection function


nickchomey:
*See [this comment](https://github.com/neuml/txtai/issues/423#issuecomment-1426777995) for details on what to implement for this issue*

First and foremost, I've been getting errors while trying to install fasttext with a different package managers (pdm and poetry) and python versions (3.8-3.10). Looking at the issues in their repo, it seems that this is a [common issue](https://github.com/facebookresearch/fastText/issues/1298) for [many people](https://github.com/python-poetry/poetry/issues/6113).

Someone created [a simple pull request](https://github.com/facebookresearch/fastText/pull/1292) 6 months ago to fix this and it has been ignored - as have hundreds of other Issues and PRs. The project just seems to be abandoned.

But [Lingua](https://github.com/pemistahl/lingua-py) appears to be a new and very promising language detection tool. There is lots of Their benchmarks show increased accuracy, though they don't have any performance metrics for comparison. 

At the very least, could the fasttext dependency be switch to [this repo ](https://github.com/cfculhane/fastText) (the 6 month old bugfix PR)?

davidmezzetti:
I came across this a few months back and was equally excited. But after investigating, the quality/speed wasn't great. While fasttext is difficult, that model works surprising well and fast. 

The bigger thing though was that it requires Python >= 3.8 while this project is >= 3.7. Python 3.7 is EOL this summer, so I would consider it again then.

I do think it would be beneficial to change the translation pipeline's langdetect parameter. If that parameter supported a callable, then someone could pass their own external language detection algorithm.

davidmezzetti:
It shouldn't fail in that case because the dependency will be satisfied. Poetry uses pip, so it should have the same dependency logic where it only tries to install if the package isn't available. If you install it beforehand, it should skip the step of installing fasttext and hence you shouldn't get an error.

davidmezzetti:
Currently, the langdetect parameter takes a path to a fasttext model. This should still be the default behavior but additional logic should be added in the `detect` method to support using an external function. A custom function would need to handle an iterable of text and return an iterable of detected languages.

The fasttext import error should only be thrown when trying to load a fasttext detector and it fails.

While implementation of external detectors **should not** be implemented, a couple ideas to test locally are:
- [Lingua](https://github.com/pemistahl/lingua-py) - Original library mentioned
- [xlm-v-base-language-id](https://huggingface.co/juliensimon/xlm-v-base-language-id)

[...]

PR #444, opened 2023-02-25
Add custom detection function for language detection

saucam:
Serves #423 

@davidmezzetti I was wondering what language ids should be returned by the custom function. Do the language ids change based on the model used? Do we need to return the same language ids as returned by the fasttext model?

Please take a look at the initial attempt.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ __pycache__/
 .coverage
 .coverage.*
 *.pyc
+.vscode/
--- a/src/python/txtai/pipeline/text/translation.py
+++ b/src/python/txtai/pipeline/text/translation.py
@@ -114,16 +114,20 @@ class Translation(HFModel):
         ids = [x.modelId for x in HfApi().list_models(author="Helsinki-NLP")] if self.findmodels else []
         return set(ids)
 
-    def detect(self, texts):
+    def detect(self, texts, custom_detect=None):
         """
         Detects the language for each element in texts.
 
         Args:
             texts: list of text
+            custom_detect: A function to use custom model to detect language
 
         Returns:
             list of languages
         """
+
+        if custom_detect is not None:
+            return custom_detect(texts)
 
         if not FASTTEXT:
             raise ImportError('Language detection is not available - install "pipeline" extra to enable')
--- a/src/python/txtai/pipeline/text/translation.py
+++ b/src/python/txtai/pipeline/text/translation.py
@@ -146,7 +146,7 @@ class Translation(HFModel):
 
         path = self.langdetect
         if not path:
-            path = "https://dl.fbaipublicfiles.com/fasttext/supervised-models/lid.176.ftz"
+            path = self.DEFAULT_LANG_DETECT
 
         if not FASTTEXT:
             raise ImportError('Language detection is not available - install "pipeline" extra to enable')
--- a/src/python/txtai/pipeline/text/translation.py
+++ b/src/python/txtai/pipeline/text/translation.py
@@ -156,7 +156,7 @@ class Translation(HFModel):
             fasttext.FastText.eprint = lambda x: None
 
             # Load language detection model
-            path = cached_download(self.langdetect, legacy_cache_layout=True)
+            path = cached_download(path, legacy_cache_layout=True)
             self.detector = fasttext.load_model(path)
 
         # Transform texts to format expected by language detection model
--- a/src/python/txtai/pipeline/text/translation.py
+++ b/src/python/txtai/pipeline/text/translation.py
@@ -26,7 +26,7 @@ class Translation(HFModel):
     # Default language detection model
     DEFAULT_LANG_DETECT = "https://dl.fbaipublicfiles.com/fasttext/supervised-models/lid.176.ftz"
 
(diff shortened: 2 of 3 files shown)


davidmezzetti:
Hi @saucam Thank you for the initial PR! 

I was thinking of reusing the langdetect parameter in the constructor:

```python
    def __init__(self, path="facebook/m2m100_418M", quantize=False, gpu=True, batch=64, langdetect=DEFAULT_LANG_DETECT, findmodels=True):
```

Then instead of adding a parameter to `detect` the function would analyze `self.langdetect` to determine if it's a function. Otherwise, it would fallback to the existing logic. Something like this.

```python
    def detect(self, texts):
        """
        Detects the language for each element in texts.

        Args:
            texts: list of text

        Returns:
            list of languages
        """

        if not FASTTEXT:
            raise ImportError('Language detection is not available - install "pipeline" extra to enable')

        if not self.detector:
            # Check if langdetect is a function
            if isinstance(self.langdetect, types.FunctionType) or hasattr(self.langdetect, "__call__"):
                self.detector = self.langdetect
            else:
                # Suppress unnecessary warning
                fasttext.FastText.eprint = lambda x: None
    
                # Load language detection model
                path = cached_download(self.langdetect, legacy_cache_layout=True)
                self.detector = fasttext.load_model(path).predict

        # Transform texts to format expected by language detection model
        texts = [x.lower().replace("\n", " ").replace("\r\n", " ") for x in texts]

        return [x[0].split("__")[-1] for x in self.detector(texts)[0]]
```

In terms of the language ids being returned, I'd leave that as an implementation detail for the custom function. Custom functions should use [ISO-639-1](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) 2 letter codes.

saucam:
While I see the advantage of re-using existing variable, I think it is overly convoluted to use a variable that can be either a function or a string!  Besides we have to add a check to find if it is a function.
Isn't the approach of having an extra variable simpler?

davidmezzetti:
I would prefer not to add another parameter. While it's easier for this change, I think the right approach is figuring out a good way to reuse langdetect. It's actually not really a useful parameter right now.

I expect there is a way to create a default detect function for fasttext and make it return two letter country codes. I'll think about it some more and write back.

saucam:
ok thank you !

saucam:
@davidmezzetti did you get a chance to give this a thought?

davidmezzetti:
Not yet, I've been tied up on other issues but I will take a look after the 5.4 release goes out.

davidmezzetti:
Sorry for the delay here.

This is the rough pseudocode to repurpose langdetect in a backwards compatible way.

Change default for langdetect to None
```python
def __init__(self, path="facebook/m2m100_418M", quantize=False, gpu=True, batch=64, langdetect=None, findmodels=True):
```

Load default detector if langdetect not provided, otherwise use external function.
```python
def detect(self, texts):
    """
    Detects the language for each element in texts.

    Args:
        texts: list of text

    Returns:
        list of languages
    """

    # Backwards compatible to load fasttext model
    if not self.langdetect or isinstance(self.langdetect, str):
        return self.defaultdetect(texts)

    # Call external language detector
    return self.langdetect(texts)
```

Default fasttext language detector
```python
def defaultdetect(self, texts):
    # Default path
    path = self.langdetect
    if not path: 
        path = "https://dl.fbaipublicfiles.com/fasttext/supervised-models/lid.176.ftz"

    if not FASTTEXT:
        raise ImportError('Language detection is not available - install "pipeline" extra to enable')

    if not self.detector:
        # Suppress unnecessary warning
        fasttext.FastText.eprint = lambda x: None

        # Load language detection model
        path = cached_download(self.langdetect, legacy_cache_layout=True)
        self.detector = fasttext.load_model(path)

    # Transform texts to format expected by language detection model
    texts = [x.lower().replace("\n", " ").replace("\r\n", " ") for x in texts]

    return [x[0].split("__")[-1] for x in self.detector.predict(texts)[0]]
```

saucam:
@davidmezzetti  thanks! I updated the PR with these changes

.gitignore
```diff
@@ -7,3 +7,4 @@ __pycache__/
 .coverage
 .coverage.*
 *.pyc
+.vscode/
```
davidmezzetti, line 10: No issue with adding this but in general I usually create my VS Code workspace outside the project.

src/python/txtai/pipeline/text/translation.py
```diff
@@ -26,7 +26,7 @@ class Translation(HFModel):
     # Default language detection model
     DEFAULT_LANG_DETECT = "https://dl.fbaipublicfiles.com/fasttext/supervised-models/lid.176.ftz"
```
davidmezzetti, line 27: Can we either delete this line or use the constant on line 149?

saucam: done

.gitignore
saucam: ok

davidmezzetti:
Thanks, I'll get this merged in once the builds done. And congratulations you're the 1,000th commit to this repo!

saucam:
Thank you! just fixed the failing tests.

davidmezzetti:
Merged, thanks for the contribution!

PR #535, opened 2023-08-28
Added cls pooling for onnx conversion

denis-gordeev:
Some models use not the mean pooling but other types of pooling. One of them is CLS-Pooling from Sentence Transformers.
For example, the current state-of-the-art model on the MTEB Leaderboard uses it (BAAI/bge-large-en).

https://huggingface.co/BAAI/bge-large-en/blob/main/1_Pooling/config.json

The code is mostly copy-pasted from the Sentence Transformers library.


davidmezzetti:
Closing as duplicate. Will be handled in #565.

PR #550, opened 2023-09-08
Align API parameter data type to translation pipeline

cstech-carl-camilleri:


davidmezzetti:
Thank you for the catch on this!

PR #666, opened 2024-02-14
Update 40_Text_to_Speech_Generation.ipynb

babinux:
waveplot doesn't seem to be working anymore.
Updated the code with a fix.

waveshow


davidmezzetti:
Thank you for this catch!

PR #704, opened 2024-05-01
Ensure config.path = None and config.path missing mean the same thing

ianawilson:
Previously if `path` was present, it was expected to have a value. This allows `path` to be `None`, and for `None` and empty to act in the same way.

Originally I made this change because setting `defaults` to `False` and `keyword` to `True` didn't seem to be enough for me.

### DIFF
--- a/src/python/txtai/embeddings/base.py
+++ b/src/python/txtai/embeddings/base.py
@@ -741,7 +741,7 @@ class Embeddings:
         self.scoring = self.createscoring() if scoring and (not isinstance(scoring, dict) or not scoring.get("terms")) else None
 
         # Dense vectors - transforms data to embeddings vectors
-        self.model = self.loadvectors() if self.config else None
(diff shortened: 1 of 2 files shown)


davidmezzetti:
Just for my clarity, the PR supports adding a a config key of `path: None`? The current method assumes path wouldn't be there but this change takes it a step further and checks that if the key is there, it has a value?

ianawilson:
> Just for my clarity, the PR supports adding a a config key of `path: None`? The current method assumes path wouldn't be there but this change takes it a step further and checks that if the key is there, it has a value?

We talked a bit [in community slack](https://txtai.slack.com/archives/C03PGPS7GMR/p1714504366020439), but the gist is that I wanted `None` and missing to act the same. The title and description have been updated for clarity.

davidmezzetti:
Thank you, appreciate the improvement!

PR #757, opened 2024-07-29
Add bm25s library to benchmarks

a0346f102085fe9f:
:wave: Hello! I've been tinkering with your benchmarking suite and added support for benchmarking bm25s which is a fast implementation of BM25 indexer and search.

This is mostly an adaptation of the code from https://www.kaggle.com/code/xhlulu/benchmark-bm25-on-beir

NDCG@10 on quora seems to line up between this and the original code, although not exactly the same.

### DIFF
--- a/examples/benchmarks.py
+++ b/examples/benchmarks.py
@@ -397,14 +397,14 @@ class BM25S(Index):
     def search(self, queries, limit):
         tokenizer = Tokenizer()
 
-        query_tokens = [tokenizer(x) for x in queries]
+        tokens = [tokenizer(x) for x in queries]
 
-        queried_results, queried_scores = self.backend.retrieve(query_tokens, corpus=self.corpus_ids, k=limit, n_threads=4)
+        results, scores = self.backend.retrieve(tokens, corpus=self.corpus_ids, k=limit, n_threads=4)
 
         # List of queries => list of matches (docid, score)
         x = []
 
-        for a, b in zip(queried_results, queried_scores):
+        for a, b in zip(results, scores):
             x.append([(str(c), float(d)) for c, d in zip(a, b)])
 
         return x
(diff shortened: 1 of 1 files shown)


davidmezzetti:
Thank you for this PR.

To ensure a fair comparison, we should use the txtai tokenizer which doesn't remove stopwords. The gist below has an example of this.

https://gist.github.com/davidmezzetti/ad5abbf36eb08694b2aa7463b8047034#file-txtai-langchain-bm25s-py-L62

a0346f102085fe9f:
Done. The scores between bm25s and scoring method are similar now (although not 100% the same). Perhaps the difference is because bm25s returns a bunch of 0-score matches when there were no matches and scoring returns none.

a0346f102085fe9f:
Ah, the checks fail. I couldn't run them on my machine because of `AttributeError: module 'pkgutil' has no attribute 'ImpImporter'.` Let me try a github codespaces machine.

davidmezzetti:
This check is with the code formatting. I'd also recommend changing the following variable names:

```
query_tokens -> tokens
queried_results -> results
queries_scores ->scores
queried_result -> result
```

I was going to do this myself after I merged the PR but since the code checks failed, figured I'd bring it up now.

a0346f102085fe9f:
Done

davidmezzetti:
One last thing. What is the reason for `del corpus_lst`? Shouldn't this automatically get deleted? 

Main reason I ask is that given this is a benchmarks scripts, it could be misleading to try to do things a normal developer wouldn't do.

a0346f102085fe9f:
I'm not sure to be honest. It was carried over from the kaggle script where it was used (I think) to ensure that there are no references held to input documents after tokenization but prior to the potentially memory intensive indexing step, which all happens inside one function.

Seeing as no other implementations in the file do that, I'll remove it.

davidmezzetti:
Thanks for putting this change together. I'll run the build again and merge it in!

Issue #814, opened 2024-11-20
Add text extraction with Docling


davidmezzetti:
[Docling](https://github.com/DS4SD/docling) looks like a promising text extraction library that could possibly augment or replace Apache Tika. 

**Update**: Docling added 3.9 support, this is a go!
~The main integration issue is that [it only supports Python 3.10+](https://github.com/DS4SD/docling/issues/385).~

yukiman76:
I started to look at this as I think it would be a great addition, but there seams to be an Library incompatibility, 

Here is some test code to show the Problem 

```Python
import logging
from txtai import Embeddings # remove this line and this will run 
# from chonkie import WordChunker, SemanticChunker, SentenceChunker, SDPMChunker
from docling.document_converter import DocumentConverter

# Configure logger
logger = logging.getLogger(__name__)
logger.setLevel(logging.DEBUG)

# Create console handler with a debug level
console_handler = logging.StreamHandler()
console_handler.setLevel(logging.DEBUG)

# Create formatter and add it to the handler
formatter = logging.Formatter('%(asctime)s - %(name)s - %(levelname)s - %(message)s')
console_handler.setFormatter(formatter)

# Add the handler to the logger
logger.addHandler(console_handler)

# Suppress lower-priority logs from the "docling" library
# logging.getLogger("docling").setLevel(logging.DEBUG)
# Configure logging for "docling"
docling_logger = logging.getLogger("docling")
docling_logger.setLevel(logging.DEBUG)  # Adjust level as needed (DEBUG, INFO, WARNING)
docling_logger.addHandler(console_handler)  # Attach the same console handler
docling_logger.propagate = True  # Ensure messages propagate to the root logger


def process_docling(sfile):
    markdown_content = ""
    try:
        logger.info("@@@@" * 10)
        logger.info(f"Processing file: {sfile}")
        docling_converter = DocumentConverter()
        logger.info("Converting file...")
        result = docling_converter.convert(sfile)
        if result is not None:
            logger.info("Exporting to markdown...")
            markdown_content = result.document.export_to_markdown()
    except Exception as e:
        logger.error(f"An error occurred: {e}")

    return markdown_content

if __name__ == "__main__":
    logger.info("Starting the script...")
    file_url = "https://www.w3.org/WAI/ER/tests/xhtml/testfiles/resources/pdf/dummy.pdf"
    logger.info(f"Input file: {file_url}")
    output = process_docling(file_url)
    logger.info(f"Output:\n{output}")
```

[...]

Issue #813, opened 2024-11-19
Embeddings.search leads to segfault error


Pringled:
Hi! When running one of the examples, I ran into an issue.

# Issue

The following code crashes with a segfault error when `search` is called:
```
UserWarning: resource_tracker: There appear to be 1 leaked semaphore objects to clean up at shutdown
  warnings.warn('resource_tracker: There appear to be %d '
```

```python
from txtai import Embeddings

# Create embeddings model, backed by sentence-transformers & transformers
embeddings = Embeddings(path="sentence-transformers/nli-mpnet-base-v2")

data = [
  "US tops 5 million confirmed virus cases",
  "Canada's last fully intact ice shelf has suddenly collapsed, " +
  "forming a Manhattan-sized iceberg",
  "Beijing mobilises invasion craft along coast as Taiwan tensions escalate",
  "The National Park Service warns against sacrificing slower friends " +
  "in a bear attack",
  "Maine man wins $1M from $25 lottery ticket",
  "Make huge profits without work, earn up to $100,000 a day"
]

# Index the list of text
embeddings.index(data)

print(f"{'Query':20} Best Match")
print("-" * 50)

# Run an embeddings search for each query
for query in ("feel good story", "climate change", "public health story", "war",
              "wildlife", "asia", "lucky", "dishonest junk"):
    # Extract uid of first result
    # search result format: (uid, score)
    uid = embeddings.search(query, 1)[0][0]

    # Print text
    print(f"{query:20} {data[uid]}")
```

# Environment info

Running on MacOS, M3, python version=3.10.14. 
Venv:
```
aiohappyeyeballs==2.4.3
aiohttp==3.11.4
aiosignal==1.3.1
annotated-types==0.7.0
anyio==4.6.2.post1
async-timeout==5.0.1
attrs==24.2.0
certifi==2024.8.30
charset-normalizer==3.4.0
click==8.1.7
diskcache==5.6.3
distro==1.9.0
exceptiongroup==1.2.2
faiss-cpu==1.9.0
fasteners==0.19
fasttext==0.9.3
filelock==3.16.1
frozenlist==1.5.0
fsspec==2024.10.0
h11==0.14.0
httpcore==1.0.7
httpx==0.27.2
huggingface-hub==0.26.2
idna==3.10
importlib_metadata==8.5.0
Jinja2==3.1.4
jiter==0.7.1
joblib==1.4.2
jsonschema==4.23.0
jsonschema-specifications==2024.10.1
[...]

Issue #744, opened 2024-07-03
Segmentation fault when calling embeddings.index


mehabakshii:
Hi there! I'm on an M1 mac and am trying to get up and running with txtai, starting with a simple script:

```
import txtai

embeddings = txtai.Embeddings()
embeddings.index(["Test", "Text"])
print(embeddings.search("positive", 1))
```

It seems like the code is getting to a seg fault on the index step - failing w/ `segmentation fault  python3 script.py`. Any idea what is going on here? I took a look at the common issues here: https://neuml.github.io/txtai/faq/#common-issues and followed steps under "Segmentation faults and similar errors on macOS" but exporting those global vars didn't work.

davidmezzetti:
Hello, thank you for writing in.

This is a nagging issue for macs that is usually solved by the FAQ solution you mentioned. 

If you wouldn't mind, what happens if you change your code to this

```python
import faiss
faiss.omp_set_num_threads(1)

import txtai

embeddings = txtai.Embeddings()
embeddings.index(["Test", "Text"])
print(embeddings.search("positive", 1))
```

Do any of the options in this work? https://github.com/facebookresearch/faiss/wiki/Threads-and-asynchronous-calls

mehabakshii:
Thanks @davidmezzetti for the suggestion! it unfortunately didn't work. 

```
(env) ➜  bond-concept python3 script.py
OMP: Error #15: Initializing libomp.dylib, but found libomp.dylib already initialized.
OMP: Hint This means that multiple copies of the OpenMP runtime have been linked into the program. That is dangerous, since it can degrade performance or cause incorrect results. The best thing to do is to ensure that only a single OpenMP runtime is linked into the process, e.g. by avoiding static linking of the OpenMP runtime in any library. As an unsafe, unsupported, undocumented workaround you can set the environment variable KMP_DUPLICATE_LIB_OK=TRUE to allow the program to continue to execute, but that may cause crashes or silently produce incorrect results. For more information, please see http://openmp.llvm.org/
[1]    96204 abort      python3 script.py
```

The suggested workaround also did not work; and still resulted in a segmentation fault. Are there other workarounds you could suggest?

mehabakshii:
Hmm I'm getting a seg fault when running `index = faiss.index_factory(data.shape[1], "IDMap,Flat")`, and same thing when ysing the `hnsw` backend. Do you recommend posting an issue on the `faiss-wheels` repo instead as it seems like this isn't specific to txtai?

[...]

Issue #717, opened 2024-05-16
Segmetation fault on macOS Sonoma 14.4.1 (MacBook Pro Apple M3 chip)


Youjin1985:
Python 3.11.7 (main, Feb  4 2024, 13:15:25) [Clang 15.0.0 (clang-1500.1.0.2.5)]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.22.2 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import txtai

In [2]: embeddings = txtai.Embeddings()

In [3]: embeddings.index(["Correct", "Not what we hoped"])
zsh: segmentation fault  ipython

Youjin1985:
Sorry, found in FAQ

PR #818, opened 2024-11-22
Macfaiss

yukiman76:
This patch addresses a known issue on macOS where FAISS can cause a segmentation fault due to a bug with OpenMP threading. Explicitly setting the number of threads to 1 is a pragmatic temporary fix while we wait for an upstream resolution.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -101,11 +101,13 @@ Segmentation faults and similar errors on macOS
 
 __Solution__
 
+
 Set the following environment parameters.
 
-- Disable OpenMP threading via the environment variable `export OMP_NUM_THREADS=1`
+- OpenMP threading is handled internally on macOS platforms, but it can be disabled by setting the environment variable OMP_NUM_THREADS to 1, e.g., export OMP_NUM_THREADS=1. For more details, refer to the related discussion on GitHub: kyamagu/faiss-wheels#100.
 - Disable PyTorch MPS device via `export PYTORCH_MPS_DISABLE=1`
 - Disable llama.cpp metal via `export LLAMA_NO_METAL=1`
+
 
 ----------
 
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -101,7 +101,6 @@ Segmentation faults and similar errors on macOS
 
 __Solution__
 
-
 Set the following environment parameters.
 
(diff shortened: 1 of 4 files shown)


src/python/txtai/ann/faiss.py
```diff
@@ -3,14 +3,22 @@
...
 from faiss import index_factory, IO_FLAG_MMAP, METRIC_INNER_PRODUCT, read_index, write_index
 from faiss import index_binary_factory, read_index_binary, write_index_binary, IndexBinaryIDMap
 
 from .base import ANN
 
+if platform.system() == "Darwin":
+    # Workaround for an open bug on macOS causing segmentation faults in FAISS.
+    # Setting the number of threads in OpenMP to 1 avoids the issue for now.
+    # Ref: https://github.com/kyamagu/faiss-wheels/issues/100
+    # TODO: Remove this workaround once the upstream bug is patched.
```
davidmezzetti, line 20: The build scripts will fail with TODO's. In general, I would reduce this comment. To a single sentence and say to refer to the txtai FAQ for more info. I'm fine handling that part but the TODO will cause a build error.

test/python/testann.py
```diff
@@ -136,6 +136,14 @@ def testNumPyLegacy(self):
         # Validate count
         self.assertEqual(ann.count(), 100)
 
+    @patch('platform.system')
+    def testmocdarwin(self, mock_platform):
```
davidmezzetti, line 140: I'd go with the existing convention (i.e. testFaissDarwin or testFaissMacOS)

docs/faq.md
```diff
@@ -101,12 +101,14 @@ Segmentation faults and similar errors on macOS
 
 __Solution__
 
+
```
davidmezzetti, line 104: Looks like some excess whitespace added. Does this link shortcut resolve to the issue?

davidmezzetti:
Thank you for the PR! Just put a few comments. Some aren't that important but the TODO one is the only one that would cause the build scripts to fail.

davidmezzetti:
This issue has been mentioned countless times. While the `export OMP_NUM_THREADS=1` workaround solves it, many won't look past the error and move on. 

Much appreciated!

References: #62 #300 #350 #463 #480 #498 #576 #685 #689 #717 #744 #813 #814

src/python/txtai/ann/faiss.py
yukiman76: Fixed

test/python/testann.py
yukiman76: Updated to testFaissMacOS

docs/faq.md
yukiman76: Updated

PR #923, opened 2025-06-28
Visualization of the txtai codebase

ivanmilevtues:
In this PR I am introducing Diagram representatioin of the txtai codebase. You can see how it renders here:
https://github.com/CodeBoarding/GeneratedOnBoardings/blob/main/txtai/on_boarding.md

I can see that you deeply care about visualizations from your readme. I think here you see things as us, visualizations really help new people to get up-to-speed. This said with this change I am adding visualization of the codebase itself, aimed at new contributors - giving them high-level understanding and allowing them to go deeper in interesting to them components. I'd love to hear your opinion on diagram first docs!

Any feedback is more than welcome! If you like what you see I can also integrate our free github action to your repo, so the diagrams are always up-to-date!

Full disclosure: we're trying to turn this into a startup, but we're still in a very early stage and figuring out what will actually be useful for people.

### DIFF
--- a/.codeboarding/AI_Pipeline_Framework.md
+++ b/.codeboarding/AI_Pipeline_Framework.md
@@ -1,0 +1,263 @@
+```mermaid
+
+graph LR
+
+    Pipeline_Factory["Pipeline Factory"]
+
+    Base_Pipeline["Base Pipeline"]
+
+    RAG_Pipeline["RAG Pipeline"]
+
+    Transcription_Pipeline["Transcription Pipeline"]
+
+    Translation_Pipeline["Translation Pipeline"]
+
+    Data_Segmentation_Pipeline["Data Segmentation Pipeline"]
+
+    Hugging_Face_Trainer_Pipeline["Hugging Face Trainer Pipeline"]
+
+    LLM_Base_Pipeline["LLM Base Pipeline"]
+
+    Hugging_Face_Model_Wrapper["Hugging Face Model Wrapper"]
+
+    Hugging_Face_Pipeline_Wrapper["Hugging Face Pipeline Wrapper"]
+
+    Pipeline_Factory -- "instantiates" --> Base_Pipeline
+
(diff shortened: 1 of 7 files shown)


davidmezzetti:
Hello. This is certainly an interesting concept, thanks for sharing!

Do you have any plans to offer a hosted way to view these diagrams? I don't think that pushing these directly into open source repositories is the best path, to be honest.

Anything that is added to an open repo eventually becomes the responsibility of the project to maintain, especially when people come and go over time. As an example, in the past I've had contributors share translations of the README. It was great initially but over time I didn't have the time or interest to maintain and ultimately deleted it when those individuals moved on to other things.

ivanmilevtues:
Thanks for the answer @davidmezzetti!

I understand your concern, I suppose that is why we introduced the github action, so that it can help keep the diagrams up-to-date without introducing extra burdon to the maintainers:
https://github.com/marketplace/actions/codeboarding-diagram-first-documentation

As said in the previous message I'd integrate it for you if it seems interesting. Apart from that we also have a VSCode extension: https://marketplace.visualstudio.com/items?itemName=Codeboarding.codeboarding

Did you have something else in mind when saying hosted i.e. in our own web-site/repository or? Let me know, as I think this way it becomes a bit harder for users to actually make use of it as it requires an extra step. Plus being in your own repository gives you the ability to customize it and own it, which in this case might not be what you are looking for.

davidmezzetti:
I guess I just don't like storing this in the main repo or being asked to maintain something like this. I'm already doing a lot of work for free, so it's just one more thing.

ivanmilevtues:
I understand @davidmezzetti! Wanted to update you that we now have a github action which will keep the documentation up-to-date with time: https://github.com/marketplace/actions/codeboarding-diagram-first-documentation

Furhter our engine now is open-source so you can see how the generation works here: https://github.com/CodeBoarding/CodeBoarding

PR #1026, opened 2026-02-15
fix: escape single quotes in dynamic column names

RinZ27:
Hardening the SQL dynamic column resolution mechanism prevents potential SQL injection when using bracket expressions.

Dynamic column names provided in SQL queries (e.g., `SELECT ... WHERE [field] = 'value'`) are translated into `json_extract` or `json_extract_string` calls. Single quotes in the column name weren't previously escaped, allowing an attacker to break out of the JSON path string literal and inject arbitrary SQL clauses.

Escaping single quotes by doubling them in both the SQLite and DuckDB backends ensures that these names are treated as literal components of the JSON path.

Regression tests have been added in `test/python/testdatabase/testsecurity.py` to cover this scenario.

Verified locally with a reproduction script and the new test suite.

### DIFF
--- a/src/python/txtai/database/sql/expression.py
+++ b/src/python/txtai/database/sql/expression.py
@@ -240,7 +240,7 @@ class Expression:
             tokens[x] = None
 
         # Set last token to resolved bracket expression
-        tokens[x] = self.resolve(self.buildtext(params), None)
+        tokens[x] = self.resolve(self.buildtext(params).replace("'", "''"), None)
 
     def similar(self, iterator, tokens, x, similar):
         """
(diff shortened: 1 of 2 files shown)


davidmezzetti:
Hello. Thank you for this PR!

While this is an interesting catch, I'd wonder what is the worst case scenario here? That someone can select all data, which they can already do with just doing a `select * from txtai`.

Nonetheless, it's good to handle. 

I believe the better place to do this would be in the SQL expression parser: https://github.com/neuml/txtai/blob/master/src/python/txtai/database/sql/expression.py#L236

Basically escape the single quotes there with double quotes so any future integrations would benefit from it.

I'd also skip adding a brand new test suite and add a test condition for this here: https://github.com/neuml/txtai/blob/master/test/python/testdatabase/testsql.py#L92

RinZ27:
@davidmezzetti Thanks for the feedback. Moving the escaping logic to the expression parser makes much more sense to ensure all backends benefit from it. 

I've updated the PR accordingly, centralized the single-quote escaping, and integrated the test case into the existing `testsql.py` as requested. Let me know if you need anything else!

PR #1037, opened 2026-02-16
Add Bayesian normalization "bb25" method for keyword scoring

### LINKED ISSUE #1023: Feature request: Bayesian BM25
Add [Bayesian BM25](https://www.researchgate.net/publication/400212695_Bayesian_BM25_A_Probabilistic_Framework_for_Hybrid_Text_and_Vector_Search), a bounded score ([0,1]) that transforms BM25 relevance scores into calibrated probability estimates. Can be useful to merge scores for hybrid search.

Implementation is inspired by this [code](https://github.com/instructkr/bb25/blob/main/src/bayesian_scorer.rs), should be something like: 
```python
import numpy as np

from .bm25 import BM25

class BayesianBM25(BM25):
    """
    Bayesian BM25 scoring.
    This class converts the standard BM25 score into a posterior probability.
    """

    def __init__(self, config=None):
        super().__init__(config)
        
        # Bayesian parameters:
        # alpha: controls the steepness of the sigmoid curve
        # beta: the raw BM25 score value that results in a 0.5 probability
        self.alpha = self.config.get("alpha", 1.0)
        self.beta = self.config.get("beta", 1.0)

    def score(self, freq, idf, length):
        """
        Calculates the Bayesian Posterior Probability for a single term.
        Matches the signature of the base BM25 class.
        """
        # 1. Calculate the standard BM25 score using the parent class method
        raw_score = super().score(freq, idf, length)
        
        if raw_score <= 0:
            return 0.0

        # 2. Likelihood: Transform raw score to probability using a Sigmoid function
        # P(Score | Relevant)
        likelihood = 1 / (1 + np.exp(-self.alpha * (raw_score - self.beta)))

        # 3. Prior: Estimate P(Relevant) based on doc structure (TF and Length)
        # Prior based on term frequency
        p_tf = 0.2 + 0.7 * min(freq / 10.0, 1.0)
        
        # Prior based on length normalization
        avgdl = self.avgdl if self.avgdl > 0 else 1.0
        ratio = length / avgdl
        p_norm = np.clip(1.0 / (1.0 + ratio), 0.1, 0.9)
        
        # Composite Prior
        prior = np.clip(0.7 * p_tf + 0.3 * p_norm, 0.1, 0.9)

        # 4. Posterior: Apply Bayes' Theorem
        # P(Relevant | Score) = (Likelihood * Prior) / Evidence
        numerator = likelihood * prior
        denominator = numerator + (1.0 - likelihood) * (1.0 - prior)
        
        return numerator / (denominator + 1e-10)
```

There is also an additional method to merge scores https://github.com/instructkr/bb25/pull/1

sigridjineth:
## Summary
- add a new `txtai.scoring.Normalize` helper to encapsulate score normalization logic
- keep current normalization as the default behavior and add a Bayesian BM25 mode via `normalize: "bayes"` / `normalize: "bb25"`
- wire TFIDF/BM25 term-search normalization through the new normalizer
- map candidate scores into `[0, 1]` with dynamic BB25 calibration (`beta=median`, `alpha_eff=alpha/std`)
- follow BB25 candidate-set behavior: estimate stats on positive-score candidates and keep zero-score candidates at `0.0`
- add/extend docs and keyword scoring tests for BB25 alias and behavior

## Notes
- this follows the issue discussion about implementing Bayesian normalization as a general normalize method for term scoring
- supports `normalize: "bayes"`, `normalize: "bb25"`, and dict config e.g. `normalize: {method: "bayes", alpha: ..., beta: ...}`
- reference implementations:
  - https://github.com/cognica-io/bayesian-bm25
  - https://github.com/instructkr/bb25
- in this generic normalize path, tf/doc-structure priors are not available, so this uses the likelihood-only BB25 variant (flat prior)

Closes #1023

### DIFF
--- a/docs/embeddings/configuration/scoring.md
+++ b/docs/embeddings/configuration/scoring.md
@@ -89,6 +89,8 @@ Enables normalized scoring (ranging from 0 to 1). This setting supports:
 - `{method: "bayes", alpha: 1.0, beta: null}` for Bayesian normalization with optional custom parameters
 
 When standard normalization is enabled, statistics from the index are used to calculate normalized scores.
+When Bayesian/BB25 normalization is enabled, it uses positive-score candidates, dynamic `beta=median(scores)`, adaptive
+`alpha_eff=alpha/std(scores)` and a sigmoid transform (likelihood-only variant with flat prior) to map scores to `[0, 1]`.
 
 Bayesian normalization references:
 - https://github.com/instructkr/bb25
--- a/src/python/txtai/scoring/normalize.py
+++ b/src/python/txtai/scoring/normalize.py
@@ -71,7 +71,7 @@ class Normalize:
 
     def bayes(self, scores):
         """
-        Bayesian normalization implementation.
+        BB25/Bayesian normalization implementation.
 
         Args:
             scores: list of (id, score)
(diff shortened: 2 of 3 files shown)


sigridjineth:
Update: aligned the BB25 normalization behavior more closely with cognica-io/bayesian-bm25 (positive-candidate stats, dynamic beta/std scaling, zero-score handling), while keeping the generic normalize-path constraints documented (likelihood-only/flat prior).

PR #1052, opened 2026-03-08
Fix translation chunk concatenation when showmodels is enabled

jnMetaCode:
## Summary

Fix incorrect tuple concatenation in `Translation.translate()` when `showmodels=True` and input text is split into multiple chunks by the tokenizer.

### Bug

When translating long text, the tokenizer may split it into multiple chunks. The `translate()` method recombines results from the same source segment using:

```python
v = (path, translated[x]) if showmodels else translated[x]
if i == last:
    results[-1] += v
```

When `showmodels=False`, `v` is a string and `+=` correctly concatenates strings.

When `showmodels=True`, `v` is a tuple `(path, text)` and `+=` concatenates tuples:
- Result: `(path, text1, path, text2)` (4 elements)
- Expected: `(path, text1 + text2)` (2 elements)

This later causes a `ValueError: too many values to unpack` when the caller does:
```python
model, op = outputs[y]
```

### Fix

Explicitly handle the `showmodels` case by concatenating only the text portion of the tuple, preserving the 2-element `(path, combined_text)` structure.

## Test plan

- [ ] Call `Translation.__call__(long_text, showmodels=True)` where `long_text` exceeds `tokenizer.model_max_length` to trigger chunk splitting
- [ ] Verify the result is `(translated_text, language, model_path)` without `ValueError`
- [ ] Verify `showmodels=False` (default) behavior is unchanged

### DIFF
--- a/test/python/testpipeline/testtext/testtranslation.py
+++ b/test/python/testpipeline/testtext/testtranslation.py
@@ -122,7 +122,7 @@ class TestTranslation(unittest.TestCase):
         # Validate no translation
         self.assertEqual(text, translation)
 
-    def testLongTranslationWithShowmodels(self):
+    def testShowmodelsChunked(self):
         """
         Test a long translation with showmodels flag. When text is chunked
         by the tokenizer, results should still be properly concatenated as


jnMetaCode:
Added `testLongTranslationWithShowmodels` in 53f0e58 — it repeats a sentence 100 times so the tokenizer splits it into chunks, then asserts the `showmodels=True` result is still a valid 3-tuple `(translation, language, model)` with `translation` as a plain string.

Let me know if you'd like any adjustments to the test.

davidmezzetti:
Please see the failed build log. I believe the test method name is too long.

Issue #217, opened 2022-02-02
Add lock to API search methods


davidmezzetti:
Currently, if an API index or upsert action is running, a parallel search action causes an error. This needs to be fixed.

davidmezzetti:
Internal performance tests showed that a full lock on the API search method actually performed better than only locking the SQL database portion of the query. While Faiss/Hnswlib are both thread-safe for search, the overall response time was better when locking happened for the full-search, once there was more than 2-3 parallel searches.

PR #1110, opened 2026-06-09
fix(database,ann): use scoped_session for per-thread SQLAlchemy sessions

kaanyalti:
## Problem

`Client` (content-store) and `PGVector` (ANN backend) both create one SQLAlchemy `Session` and hold it for the engine's lifetime. `StaticPool` means every `engine.connect()` call returns the same underlying DBAPI connection, so all threads share one Session object.

SQLAlchemy's `Session` is not thread-safe. Once any thread leaves the Session's internal state machine in `'prepared'` state, for example when an uncommitted read is interrupted or `_commit_read_txns()` races a write, every subsequent operation raises `InvalidRequestError: This Session's transaction has been rolled back due to a previous exception`. That error persists until the process restarts.

Issue #217 ("Add lock to API search methods") added a per-method lock to work around this. This PR fixes the source of the problem.

## Fix

**`txtai/database/client.py`**

- Drops `StaticPool`, uses the default `QueuePool`. Each thread holds at most one connection at a time; idle connections return to the pool between operations.
- `connect()` returns `scoped_session(sessionmaker(bind=engine))` instead of a bare `Session`. The proxy routes any attribute access to the calling thread's own Session instance automatically.
- `Batch` and `Score` are `TEMPORARY TABLE`s, which are connection-scoped in SQLite and Postgres. A `connect` event (`_init_temp_tables`) creates them on each new pool connection, so every thread's Session gets its own temp workspace. No cross-thread interference on `DELETE FROM batch`.
- Persistent-table DDL (`documents`, `sections`, `objects`, reindex scratch) goes through `engine.begin()` or `CreateTable` via the current Session, so DDL is committed before other sessions need to see it, and no second connection competes with an existing write lock.

**`txtai/ann/dense/pgvector.py`**

- Same `scoped_session(sessionmaker)` change for `self.database`.
- DDL (`CREATE TABLE vectors`, index create/drop) uses a new `sqldialect_conn()` helper that runs inside an explicit `engine.begin()` connection rather than the scoped Session. DDL commits immediately.
- `save()` calls only `self.database.commit()`. The separate `self.connection.commit()` is removed because there is no longer a separate DDL connection.

## Validation

`testConcurrentSearch` (added to `test/python/testdatabase/testclient.py`): indexes data with a `Client`-backed `Embeddings` instance, commits the transaction, then runs 8 search threads at the same time. Checks that no thread raised an error and that at least one returned the correct result.

Full `TestClient` suite: 42 passed, 1 skipped (`testSparse`, pre-existing failure from a missing optional dependency, unrelated). No regressions against the 41/41 baseline.

pgvector tests require a live Postgres server and could not run locally. The DDL-via-`engine.begin()` approach is documented in SQLAlchemy's multi-session setup guides for auto-committed DDL.

## Prior art

Issue #217 patched the symptom with an RLock around search methods. SQLAlchemy's own docs recommend `scoped_session` for multi-threaded code where an ORM object is shared across threads. This PR applies that pattern, which means the lock in #217 is no longer needed.

### DIFF
--- a/src/python/txtai/ann/dense/pgvector.py
+++ b/src/python/txtai/ann/dense/pgvector.py
@@ -8,8 +8,8 @@ import os
 try:
     from pgvector.sqlalchemy import BIT, HALFVEC, VECTOR
 
-    from sqlalchemy import create_engine, delete, func, text, Column, Index, Integer, MetaData, StaticPool, Table
-    from sqlalchemy.orm import Session
+    from sqlalchemy import NullPool, create_engine, delete, func, text, Column, Index, Integer, MetaData, Table
+    from sqlalchemy.orm import Session, scoped_session, sessionmaker
     from sqlalchemy.schema import CreateSchema
 
     PGVECTOR = True
@@ -37,7 +37,7 @@ class PGVector(ANN):
             raise ImportError('PGVector is not available - install "ann" extra to enable')
 
         # Database connection
-        self.engine, self.database, self.connection, self.table = None, None, None, None
+        self.engine, self.database, self.table = None, None, None
 
         # Scalar quantization
         quantize = self.config.get("quantize")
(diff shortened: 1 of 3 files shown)


davidmezzetti:
AI Slop.

PR #1114, opened 2026-06-16
fix(tokenizer): respect lowercase=False in ngramtokenize

JSap0914:
## Bug

`ngramtokenize` called `text.lower()` unconditionally, silently overriding `lowercase=False`. `__call__` already lowercases the input when `self.lowercase is True`, so the extra `.lower()` inside `ngramtokenize` was both redundant for the default case and wrong when `lowercase=False`.

```python
tokenizer = Tokenizer(lowercase=False, ngrams=3)
tokenizer('CAT')   # returned ['cat']  ← wrong; should be ['CAT']
```

## Fix

Remove the redundant `.lower()` from line 165 of `tokenizer.py` so that case handling is governed exclusively by `self.lowercase` in `__call__`, consistent with every other tokenization path (whitespace, regexp, segment).

## Verification

```
pytest test/python/testpipeline/testdata/testtokenizer.py -v
```

All 4 tests pass, including the new `testNgramLowercase` regression test.

> AI-assisted contribution

### DIFF
--- a/src/python/txtai/pipeline/data/tokenizer.py
+++ b/src/python/txtai/pipeline/data/tokenizer.py
@@ -162,7 +162,7 @@ class Tokenizer(Pipeline):
         edge = self.ngrams.get("edge", False)
 
         # Split on non-whitespace and apply optional word padding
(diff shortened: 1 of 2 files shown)


davidmezzetti:
Thank you for the fix. The build is failing due to some other reason.

Issue #1115, opened 2026-06-16
FastAPI 0.137+ modified how routers work


davidmezzetti:
[FastAPI 0.137](https://github.com/fastapi/fastapi/releases/tag/0.137.0) introduced breaking changes for `txtai`. The current custom routing class ignores injected dependencies. 

This fix will address that moving forward. Anyone who runs into this issue should stay on FastAPI <= 0.136.1 until `txtai 9.11` is released.

PR #1116, opened 2026-06-16
Fix EncodingAPIRoute to honour app-level dependencies via FastAPI effective route context

JSap0914:
## Bug

`testapi.testauthorization.TestAuthorization.testInvalid` has been failing on master: an unauthenticated `GET /search?query=test` returns `200` instead of the expected `401`.

## Root cause

FastAPI ≥ 0.94 propagates app-level and router-level dependencies to included routes through a `_effective_route_context_var` `ContextVar`.  The flow in `APIRoute.handle` is:

```python
token = _effective_route_context_var.set(effective_context)   # ← set
try:
    app = request_response(self.get_route_handler())           # ← get_route_handler called here
finally:
    _effective_route_context_var.reset(token)                  # ← reset BEFORE handler runs
await app(scope, receive, send)
```

`EncodingAPIRoute.get_route_handler()` overrides the base implementation to pick a per-request response class via `ResponseFactory`.  The override called `get_request_handler()` with `self.dependant` (the bare route dependant, which contains only per-route dependencies).  It never read `_effective_route_context_var`, so the app-level `Authorization` dependency was silently dropped for every route served by this router class.

Note: reading the ContextVar *inside* the inner `async def handler(request)` closure does not work because by the time that coroutine is awaited the ContextVar has already been reset in the `finally` block above.

## Fix

Read `_effective_route_context_var` **synchronously at the top of `get_route_handler()`**, while the ContextVar is still set.  Capture the result in the closure so that `get_request_handler()` receives the effective-context `dependant` (which contains the combined app + router + route dependencies) rather than the bare route dependant.

When no effective context is present (e.g. the initial startup call that creates `self.app`) the behaviour is unchanged: `self` is used as before.

Changed file: `src/python/txtai/api/route.py`

## Verification

```
# RED (before fix)
python -m unittest testapi.testauthorization -v
# testInvalid FAIL: 200 != 401

# GREEN (after fix)
python -m unittest testapi.testauthorization -v
# testInvalid ... ok
# testValid   ... ok
```

Broader `testapi.testextension` tests also pass.  Pre-existing failures in `testencoding` (`testMessagePack`, `testImages`, `testReadOnly`) are unrelated to this change and reproduce identically on unmodified master.

### DIFF
--- a/src/python/txtai/api/route.py
+++ b/src/python/txtai/api/route.py
@@ -2,7 +2,7 @@
 Route module
 """
 
-from fastapi.routing import APIRoute, get_request_handler
+from fastapi.routing import APIRoute, _effective_route_context_var, get_request_handler
 
 from .responses import ResponseFactory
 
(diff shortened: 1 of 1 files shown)


davidmezzetti:
Already fixed with #1115

PR #1121, opened 2026-06-25
Add TwelveLabs Marengo embeddings and Pegasus video pipeline

mohit-twelvelabs:
Hi! I'm Mohit, I work at TwelveLabs (@mohit-twelvelabs).

This PR adds an opt-in [TwelveLabs](https://twelvelabs.io) integration with two pieces:

- **Marengo multimodal embeddings backend** (`txtai/vectors/dense/twelvelabs.py`) — a new dense vectors method that builds 512-dim embeddings in a shared text/image/audio/video space via the TwelveLabs API. It plugs into `VectorsFactory` like the existing `litellm`/`model2vec` backends and is inferred from a `path` starting with `marengo` (e.g. `marengo3.0`). Text inputs are strings; other modalities are dicts (`{"image_url": ...}`, `{"audio_url": ...}`), enabling cross-modal similarity search.
- **Pegasus video pipeline** (`txtai/pipeline/video/pegasus.py`) — a new `Pegasus` pipeline that analyzes videos (public url or TwelveLabs asset id) and generates text given a prompt: summarize, caption, Q&A, structured extraction. It follows the standard `Pipeline.__call__` contract and is auto-registered by `PipelineFactory`, so it works in Python, YAML config, workflows and the API.

**Why it helps txtai:** Marengo brings native multimodal (video/audio/image) embeddings to txtai's existing embeddings index, and Pegasus adds video understanding to the pipeline/workflow system — both areas txtai doesn't currently cover with a hosted model.

**Opt-in / non-breaking:** No defaults change. Both components are reached only when you explicitly select `marengo*` as a vectors path or instantiate the `pegasus` pipeline. The `twelvelabs>=1.2.8` SDK is added to the existing `vectors` and `pipeline-llm` extras (conditional import, with a clear ImportError pointing at the extra). API keys are read from `TWELVELABS_API_KEY` or config.

**Testing:**
- No-network unit tests for both (mock the SDK client): `test/python/testvectors/testdense/testtwelvelabs.py` and `test/python/testpipeline/testvideo/testpegasus.py`, mirroring the existing litellm test style.
- Key-gated live tests (skip without `TWELVELABS_API_KEY`). The Marengo live test passes end-to-end (real 512-dim vector through the indexing path). The Pegasus pipeline wiring is verified against the live API (request is correctly built, authenticated, and accepted into the analyze endpoint); a full transcript run depends on a server-side-valid sample video.
- `black --line-length 150` clean; `pylint --rcfile=.pylintrc` 10.00/10 on the new source modules.

You can grab a free API key at https://twelvelabs.io — there's a generous free tier.

### DIFF
--- a/docs/embeddings/configuration/vectors.md
+++ b/docs/embeddings/configuration/vectors.md
@@ -12,13 +12,13 @@ Sets the path for a vectors model. When using a transformers/sentence-transforme
 
 ## method
 ```yaml
-method: transformers|sentence-transformers|llama.cpp|litellm|model2vec|external|
-        words
+method: transformers|sentence-transformers|llama.cpp|litellm|model2vec|twelvelabs|
+        external|words
 ```
 
 Embeddings method to use. If the method is not provided, it is inferred using the `path`.
 
-`sentence-transformers`, `llama.cpp`, `litellm`, `model2vec` and `words` require the [vectors](../../../install/#vectors) extras package to be installed.
+`sentence-transformers`, `llama.cpp`, `litellm`, `model2vec`, `twelvelabs` and `words` require the [vectors](../../../install/#vectors) extras package to be installed.
 
 ### transformers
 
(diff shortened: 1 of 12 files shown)


davidmezzetti:
Hi Mohit. Thank you for this contribution. Would it be possible for you to do this as a third-party library that you host? I don't integrate proprietary APIs into `txtai`.

Third party plugins are supported with the `vectors` and `pipeline` package. All you need to do is pass the full class. (i.e. twelvelabs.txtai.vectors.TwelveLabs etc).

mohit-twelvelabs:
Thanks for the quick and clear response, @davidmezzetti — completely understand not pulling proprietary APIs into core. The third-party `vectors`/`pipeline` class hook is a perfect fit: I'll publish it as a standalone package that plugs in via the full class path (e.g. `twelvelabs.txtai.vectors.TwelveLabs`) and link to it from our docs. Closing this in favor of that approach. Thanks again for the pointer!

— Mohit (@mohit-twelvelabs, TwelveLabs)

davidmezzetti:
Sounds good. If you end up doing this, let me know. I'm happy to share on socials.

PR #1124, opened 2026-07-01
feat: add pandas DataFrame support for Tabular pipeline and Embeddings

Muhammad-Ikhwan-Fathulloh:
## Summary

This PR adds native support for **pandas DataFrame** inputs to both the **Tabular pipeline** and the **Embeddings** `index()` and `upsert()` methods. This allows users to pass DataFrames directly without manually converting them into lists of dictionaries.

## Changes

- Added DataFrame detection and processing in the **Tabular** pipeline.
- Added DataFrame support in **Embeddings.index()** and **Embeddings.upsert()**.
- Automatically converts DataFrame rows into record dictionaries for indexing and upserting.
- Added conditional `pandas` imports using `try-except ImportError` to keep pandas as an optional dependency.
- Added unit tests to verify DataFrame support while ensuring existing functionality remains unchanged.

### Embeddings

```python
import pandas as pd
from txtai.embeddings import Embeddings

df = pd.DataFrame([
    {"id": 0, "text": "This is a test document"},
    {"id": 1, "text": "Another test document"},
])

embeddings = Embeddings({
    "path": "sentence-transformers/all-MiniLM-L6-v2"
})

# Index directly from a DataFrame
embeddings.index(df)

# Upsert from a DataFrame
embeddings.upsert(df)

# Search
print(embeddings.search("test"))
```

### Tabular Pipeline

```python
import pandas as pd
from txtai.pipeline import Tabular

df = pd.DataFrame([
    {"text": "This is a test document"},
    {"text": "Another test document"},
])

tabular = Tabular()

for row in tabular(df):
    print(row)
```

## Notes

- This is a non-breaking change.
- Existing input types (lists, generators, iterables, etc.) continue to work without modification.
- The implementation gracefully handles environments where `pandas` is not installed by using conditional imports.

### DIFF
--- a/src/python/txtai/embeddings/index/stream.py
+++ b/src/python/txtai/embeddings/index/stream.py
@@ -2,6 +2,13 @@
 Stream module
 """
 
+# Conditional import for pandas
+try:
+    import pandas as pd
+    PANDAS = True
+except ImportError:
+    PANDAS = False
+
 from .autoid import AutoId
 from .transform import Action
 
@@ -41,6 +48,10 @@ class Stream:
         Args:
             documents: input documents
         """
+
+        # Handle pandas DataFrame input
+        if PANDAS and isinstance(documents, pd.DataFrame):
+            documents = documents.to_dict("records")
 
         # Iterate over documents and yield standard (id, data, tag) tuples
         for document in documents:
--- a/src/python/txtai/pipeline/data/tabular.py
+++ b/src/python/txtai/pipeline/data/tabular.py
@@ -68,13 +68,17 @@ class Tabular(Pipeline):
                     raise ValueError(f"Invalid local file path or file not found: {item}")
 
             # Dict
-            if isinstance(item, dict):
+            elif isinstance(item, dict):
                 dicts.append(item)
 
             # List of dicts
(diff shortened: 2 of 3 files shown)


davidmezzetti:
Hello. Thank you for this PR. While the tabular change seems like it could be useful I don't see much value in adding a new (even if it's optional) dependency for the core framework for that one liner. You can also do the same with polars in a one liner and it doesn't seem worth it adding that in either.

Muhammad-Ikhwan-Fathulloh:
> Hello. Thank you for this PR. While the tabular change seems like it could be useful I don't see much value in adding a new (even if it's optional) dependency for the core framework for that one liner. You can also do the same with polars in a one liner and it doesn't seem worth it adding that in either.

Ah, I see. That makes sense thanks for the explanation!

I understand your point about keeping the core framework minimal and avoiding framework-specific handling when users can achieve the same result with a simple one-liner. I hadn't fully considered the long-term maintenance and consistency aspects from that perspective.

I'll go ahead and close this PR. Going forward, I'll keep this design philosophy in mind for future contributions and try to align them more closely with the project's direction.

Thanks again for the feedback!

PR #1129, opened 2026-07-12
Add Milvus ANN backend

zc277584121:
## Summary
- Add a dense Milvus ANN backend using `pymilvus.MilvusClient` with an explicit `indexid` primary key and `embedding` FLOAT_VECTOR schema.
- Register `backend: milvus`, add `pymilvus[milvus-lite]>=3.0.0` to the `ann` extra, and document Milvus configuration.
- Add a real Milvus Lite test covering factory creation, index, append, search, save/load, count, and delete.

## Usage
```yaml
backend: milvus
milvus:
    uri: ./milvus.db
    collection: vectors
```

Set `MILVUS_URI` and optionally `MILVUS_TOKEN` to connect to Milvus server or Zilliz Cloud.

## Tests
- `MINIMAL=1 uv pip install -e . pytest numpy 'pymilvus[milvus-lite]>=3.0.0'`
- `.venv/bin/python -m py_compile src/python/txtai/ann/dense/milvus.py src/python/txtai/ann/dense/factory.py src/python/txtai/ann/dense/__init__.py test/python/testann/testdense.py`
- `.venv/bin/python -m pytest test/python/testann/testdense.py -k Milvus -q`
- `black --check src/python/txtai/ann/dense/milvus.py src/python/txtai/ann/dense/factory.py src/python/txtai/ann/dense/__init__.py test/python/testann/testdense.py`
- `git diff --check upstream/master...HEAD`

### DIFF
--- a/docs/embeddings/configuration/ann.md
+++ b/docs/embeddings/configuration/ann.md
@@ -4,7 +4,7 @@ Approximate Nearest Neighbor (ANN) index configuration for storing vector embedd
 
 ## backend
 ```yaml
-backend: faiss|hnsw|annoy|ggml|numpy|torch|turbovec|zvec|pgvector|sqlite|custom
+backend: faiss|hnsw|annoy|ggml|numpy|torch|turbovec|zvec|milvus|pgvector|sqlite|custom
 ```
 
 Sets the ANN backend. Defaults to `faiss`. Additional backends are available via the [ann](../../../install/#ann) extras package. Set custom backends via setting this parameter to the fully resolvable class string.
@@ -116,6 +116,12 @@ zvec:
 ```
 
 The [zvec](https://github.com/alibaba/zvec) backend is an embedded, path-based vector index.
+
+### milvus
+
+The [Milvus](https://milvus.io/docs/milvus_lite.md) backend uses Milvus Lite as an embedded, path-based vector index. It does not require a running Milvus server.
+
+Milvus indexes vectors with `AUTOINDEX` and inner product. Inner product is equivalent to cosine similarity on normalized vectors.
 
 ### pgvector
 ```yaml
--- a/setup.py
+++ b/setup.py
@@ -60,6 +60,7 @@ extras["ann"] = [
     "ggml-python>=0.0.41",
     "hnswlib>=0.5.0",
     "pgvector>=0.4.1",
+    "pymilvus[milvus-lite]>=3.0.0",
     "scikit-learn>=0.23.1",
     "scipy>=1.4.1",
     "sqlalchemy>=2.0.20",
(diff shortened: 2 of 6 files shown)


davidmezzetti:
Hello, thanks for this PR!

To date, I've been hesitant to add pure vector stores to `txtai` and have focused more on local vector indexes. For pure vector stores, I've leaned more on having a separate plugin like this: https://github.com/qdrant/qdrant-txtai

davidmezzetti:
Perhaps this makes more sense with this: https://github.com/milvus-io/milvus-lite?

davidmezzetti:
Perhaps @morgan-coded would be interested in doing a milvus-lite version of this instead? Seems like it would be similar to the zvec one.

morgan-coded:
Happy to help if there’s still anything you’d like me to take on here. Let me know.

davidmezzetti:
Wouldn't it make sense to add `milvus-lite` to setup.py instead?

morgan-coded:
Glad to build the `milvus-lite` ANN backend. I’ll keep it embedded around a local `.db` file, mirror the zvec backend in #1130, and add `milvus-lite` to `setup.py`. I’m planning to open that as a separate PR so it doesn’t cut across the work already in #1129; is that how you’d like it handled, or would you rather steer #1129 itself toward the lite version? I’m assuming embedded-only is the intended scope, with no remote Milvus server or Zilliz Cloud path.

chuenchen309 (commented):
Read through this against the existing `faiss`/`pgvector` backends — it's clean, and the id/offset semantics and `metadata()`/`self.backend` wiring line up with the base `ANN` contract. Two small things:

**`count()` and flush timing.** `append()`/`delete()` mutate the collection, but only `save()` calls `flush()`. `runTests` asserts exact counts right after `append()`/`delete()` with no intervening save, so `get_collection_stats["row_count"]` has to reflect un-flushed writes. That holds on Milvus Lite here, but since it's a timing-sensitive guarantee, would an explicit `flush()` in `count()` (or after `insert`/`delete`) make it more robust — or is the Lite-only behavior intentional? Just want to make sure the count assertions aren't leaning on segment-flush timing.

**Dependency floor.** `pymilvus[milvus-lite]>=3.0.0` pins to the just-released 3.0 major and excludes the 2.5/2.6 line many installs currently resolve to. The `MilvusClient` APIs used here (`create_schema`, `prepare_index_params`, `AUTOINDEX`) have been stable since 2.4 — is 3.0 required for milvus-lite compatibility, or could the floor be lower to ease resolution for existing users?

One thing I'd keep as-is: the `validate()` schema check on load is a nice defensive touch that the other backends don't have.

Nice addition overall — Milvus Lite as an embedded, path-based index fits the `zvec`/`sqlite`/`pgvector` family well.

davidmezzetti:
The only recommendation here is to use the `milvus-lite` dependency instead of `pymilvus`.

PR #1131, opened 2026-07-12
Fix a small issue in txtai

bglglzd:
This is a quick cleanup of whitespace and line endings where they were inconsistent.


davidmezzetti:
Closes as duplicate of #1132

PR #1136, opened 2026-07-14
Fix Graph.insert() node id desync when a document's text/object field is None

chuenchen309:
## Summary

`Graph.insert()` decided whether a document consumed an `indexid` slot based on whether the resolved text/object value was **not `None`**. But the actual vector index offset, computed by `Transform.stream()` (`embeddings/index/transform.py`), increments based on **key presence** (`self.text in document[1]`), regardless of the value.

A document like `{"text": None}` consumes a real vector index slot (per `Transform.stream()`'s criterion) but was silently skipped by `Graph.insert()` without incrementing its own `index` counter — so every subsequent document in the batch got assigned to the wrong graph node id. This is a silent, permanent misattribution of node ids to the wrong document data whenever any indexed document has a `text`/`object` key present with value `None` (plausible with DB/pandas-sourced documents with nullable columns).

```python
graph.insert([
    ("d0", {"text": "alpha content"}, None),
    ("d1", {"text": None}, None),
    ("d2", {"text": "gamma content"}, None),
], index=0)
# before this fix: node 0 = d0, node 1 = d2 (wrong! d2's actual vector-index slot is 2, not 1)
# after this fix:  node 0 = d0, node 2 = d2 (node 1 correctly left empty - d1 consumed that slot)
```

## Fix

Track slot consumption with the same key-presence criterion `Transform.stream()` uses (`hasfield`), separately from whether a node actually gets created for that slot (still gated on the resolved value being non-`None`, so no node is created for a genuinely empty document — but the index still advances, leaving a gap instead of shifting every later id).

## Testing

Added `testInsertNoneTextIndexSync` to `test/python/testgraph.py`. Confirmed via `git stash` that it fails against the pre-fix code (`d2` lands at node id 1 instead of 2) and passes after.

- Ran the affected test plus `testEdges`/`testCustomBackend`/`testCustomBackendNotFound` — all pass, no regressions.
- `pylint` (10.00/10) clean on the changed files.

## AI disclosure

Developed with AI assistance (Claude Code), which located the mismatch by comparing `Graph.insert()`'s slot-consumption criterion against `Transform.stream()`'s offset-computation criterion. I reviewed the diff, independently verified the bug and the fix by executing both against the actual `Graph.insert()` method, and ran the tests/linters myself before opening this PR.

### DIFF
--- a/src/python/txtai/graph/base.py
+++ b/src/python/txtai/graph/base.py
@@ -409,6 +409,12 @@ class Graph:
             # Manually provided relationships and attributes to copy
             relations, attributes = None, {}
 
+            # Tracks whether this document consumes an indexid slot. Must match the same
+            # key-presence criterion Transform.stream() uses to compute its offset, so node
+            # ids stay in sync with the real vector index even when a text/object field is
+            # present but set to None.
+            hasfield = True
+
             # Extract data from dictionary
             if isinstance(document, dict):
                 # Extract relationships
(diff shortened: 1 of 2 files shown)


davidmezzetti:
To follow up, I researched this a good deal and this change isn't necessary. If it was it would affect more than the graph. It would also effect databases, scoring indexes and more.

I can see why AI would get tripped up on this though. Yes it does yield an indexable item in this case but ANNs and Scoring instances either fail or skip `None` values.

If you can find an example that fails via the `Embeddings` interface, I'll reevaluate.

PR #1137, opened 2026-07-14
Fix Graph.filter() crash on ids that don't exist in the graph

chuenchen309:
## Summary

`Graph.filter()` unpacked `self.node(node)` with `**` into `addnode()` for every input id, but `self.node()` returns `None` for an id not present in the graph. `**None` raises `TypeError` instead of the filter simply skipping an id it doesn't recognize — the docstring says `nodes` are "nodes to select", with no stated precondition that every id must already exist.

```python
graph.addnode(0, id="a", data="hello")
graph.addnode(1, id="b", data="world")
graph.filter([0, 1, 99])
# TypeError: NetworkX.addnode() argument after ** must be a mapping, not NoneType
```

A realistic trigger for this: a caller-supplied id list that includes a stale/deleted id, or an id from a different (mismatched) index.

## Fix

Skip ids where `self.node()` returns `None` instead of crashing.

## Testing

Added `testFilterUnknownId` to `test/python/testgraph.py`. Confirmed via `git stash` that it fails with the pre-fix `TypeError` and passes after; the known-good ids and their edge are still correctly included in the filtered subgraph.

- Ran the affected test plus `testInsertNoneTextIndexSync`/`testEdges` — no regressions.
- `pylint` (10.00/10) clean on the changed files.

## AI disclosure

Developed with AI assistance (Claude Code), which located the crash while investigating the `graph/base.py` module for defects, tracing `self.node()`'s `None` return contract against its unguarded use in `filter()`. I reviewed the diff, independently reproduced the crash and the fix by executing both against the actual method, and ran the tests/linters myself before opening this PR.

### DIFF
--- a/src/python/txtai/graph/base.py
+++ b/src/python/txtai/graph/base.py
@@ -544,8 +544,13 @@ class Graph:
             # Unpack node and score, if available
             node, score = node if isinstance(node, tuple) else (node, None)
 
+            # Skip ids that don't exist in this graph
+            nodeattrs = self.node(node)
+            if nodeattrs is None:
+                continue
+
             # Add nodes
-            graph.addnode(node, **self.node(node))
+            graph.addnode(node, **nodeattrs)
 
             # Add score if present
             if score is not None:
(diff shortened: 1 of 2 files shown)


davidmezzetti:
While I appreciate these PRs, I ask that you space it out. I'm not sure these graph issues are really valid or they are covering very uncommon use cases but it's taking a lot of time to validate.

davidmezzetti:
Thank you for this PR. I have a bit of cleanup to do but it's a good change.

PR #1142, opened 2026-07-16
Fix Terms KeyError on a query with no similar() clause

chuenchen309:
`Embeddings.terms()`/`batchterms()` read `parse["similar"]` unconditionally, but `SQL.__call__` only adds the `similar` key when the query has a `similar()` clause. A keyword-free SQL query (e.g. `select id, text from txtai where id = 1`) therefore raised `KeyError: 'similar'` instead of returning an empty term string. Other consumers of that key (`explain`, `scan`) already guard it with `if "similar" in query`.

Fix: `parse.get("similar", [])`. Added `testTerms` covering the keyword, similar-clause, and no-similar-clause paths.

Prepared with AI assistance and reviewed before submission.

### DIFF
--- a/src/python/txtai/embeddings/search/terms.py
+++ b/src/python/txtai/embeddings/search/terms.py
@@ -37,8 +37,8 @@ class Terms:
                 # Parse query
                 parse = self.database.parse(query)
 
-                # Join terms from similar clauses
-                terms.append(" ".join(" ".join(s) for s in parse["similar"]))
+                # Join terms from similar clauses (absent when the query has no similar() clause)
+                terms.append(" ".join(" ".join(s) for s in parse.get("similar", [])))
 
             return terms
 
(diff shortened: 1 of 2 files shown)


src/python/txtai/embeddings/search/terms.py
```diff
@@ -37,8 +37,8 @@ def __call__(self, queries):
                 # Parse query
                 parse = self.database.parse(query)
 
-                # Join terms from similar clauses
-                terms.append(" ".join(" ".join(s) for s in parse["similar"]))
+                # Join terms from similar clauses (absent when the query has no similar() clause)
+                terms.append(" ".join(" ".join(s) for s in parse.get("similar", [])))
```
davidmezzetti, line 41: It would make more sense to check if similar in parse before adding terms instead of adding empty strings.

This would require also changing embeddings.terms as follows:

```python
        results = self.batchterms([query])
        return results[0] if results else results
```

This is similar logic for explain / batchexplain

davidmezzetti:
Please make sure tests are put in the proper place. Most of the classes have tests in alphabetical order.

PR #1144, opened 2026-07-16
Fix msgpack Accept header crashing streaming API endpoints

chuenchen309:
`EncodingAPIRoute` re-wraps a response as `MessagePackResponse` by reading `response.body`, but a `StreamingResponse` has no `.body`. A client sending `Accept: application/msgpack` to a streaming endpoint (e.g. streaming LLM/chat completions, transcription) therefore hit `AttributeError` → HTTP 500. Both msgpack responses and streaming are first-class features, so the combination is legal input.

Fix: skip the msgpack re-wrap for `StreamingResponse`, passing the stream through unchanged (same as the non-msgpack path). Added `testMessagePackStreaming` (minimal app with a streaming endpoint); it fails before the fix and passes after.

Prepared with AI assistance and reviewed before submission.

### DIFF
--- a/test/python/testapi/testencoding.py
+++ b/test/python/testapi/testencoding.py
@@ -121,6 +121,34 @@ class TestEncoding(unittest.TestCase):
         # Test reading image
         self.assertIsInstance(PIL.Image.open(BytesIO(results[0]["object"])), PIL.Image.Image)
 
+    def testMessagePackStreaming(self):
+        """
+        Test a streaming response is passed through unchanged when msgpack is requested
+        """
+
+        from fastapi import FastAPI
+        from fastapi.responses import StreamingResponse
+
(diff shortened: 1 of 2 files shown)


test/python/testapi/testencoding.py
```diff
@@ -121,6 +121,34 @@ def testMessagePack(self):
         # Test reading image
         self.assertIsInstance(PIL.Image.open(BytesIO(results[0]["object"])), PIL.Image.Image)
 
+    def testMessagePackStreaming(self):
+        """
+        Test a streaming response is passed through unchanged when msgpack is requested
+        """
+
+        from fastapi import FastAPI
```
davidmezzetti, line 129: Imports should be at the top not inline in a function.

PR #1149, opened 2026-07-16
Add milvus-lite embedded dense ANN backend

morgan-coded:
Introduces a `milvus-lite` embedded ANN backend, following the thread in #1129 — separate PR, dense-only, mirroring the zvec backend's shape.

Credit to @zc277584121 for opening the lane in #1129 — this PR follows the maintainer's milvus-lite recommendation from that thread.

Scope per your recommendations there: depends on `milvus-lite` directly (no pymilvus), added to the `ann` extra in setup.py, pinned `>=3.1` — the 3.1.0 line is pure-Python, so the ubuntu/macos/windows CI matrix stays clean with no platform guards. Lifecycle matches zvec: close-before-reopen on load/index, save via ArchiveFactory, close() releases engine handles before removing temp state.

One trade-off stated up front: milvus-lite 3.x documents the native `MilvusLite` embedded API as its internal engine (their packaged public path is `pymilvus[milvus-lite]`). Using it directly matches your recommendation and keeps the dependency light; the backend is deliberately thin so any drift in a future minor is cheap to absorb — happy to own that maintenance.

Tests mirror testZvec/testZvecCustom: index/search round-trip, save/load, close/lock-release, and a custom-config case.

### DIFF
--- a/docs/embeddings/configuration/ann.md
+++ b/docs/embeddings/configuration/ann.md
@@ -4,7 +4,7 @@ Approximate Nearest Neighbor (ANN) index configuration for storing vector embedd
 
 ## backend
 ```yaml
-backend: faiss|hnsw|annoy|ggml|numpy|torch|turbovec|zvec|pgvector|sqlite|custom
+backend: faiss|hnsw|annoy|ggml|milvuslite|numpy|torch|turbovec|zvec|pgvector|sqlite|custom
 ```
 
 Sets the ANN backend. Defaults to `faiss`. Additional backends are available via the [ann](../../../install/#ann) extras package. Set custom backends via setting this parameter to the fully resolvable class string.
--- a/docs/embeddings/configuration/ann.md
+++ b/docs/embeddings/configuration/ann.md
@@ -4,7 +4,7 @@ Approximate Nearest Neighbor (ANN) index configuration for storing vector embedd
 
 ## backend
 ```yaml
-backend: faiss|hnsw|annoy|ggml|milvuslite|numpy|torch|turbovec|zvec|pgvector|sqlite|custom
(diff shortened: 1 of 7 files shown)


davidmezzetti:
Sorry for the confusion, I wasn't thinking of having both #1129 and this. But just one PR that adds Milvus only using Milvus light. I would even just call it Milvus. I should have been more clear.

morgan-coded:
Renaming it to plain `milvus` works — that's a small push here (backend key, factory, extra, docs, tests all follow). And to keep it to one PR per your note: whichever shape you prefer — this PR renamed, or the work folded under #1129 — I'm good either way; zc277584121's #1129 kicked the whole lane off regardless. Say the word and it's done.

zc277584121:
@davidmezzetti @morgan-coded I am aligned with the single-PR Milvus Lite direction here.

A core backend named `milvus`, implemented with Milvus Lite only, makes sense for txtai because it keeps the integration in the same embedded/local ANN family as `zvec` and avoids turning this into a remote vector store integration.

I am happy for this PR to be the main path once it is renamed from `milvuslite` to `milvus` across the backend key, factory, docs, tests, and config examples. After that, I can close #1129 as superseded so there is only one Milvus PR under review.

Thanks for picking this up, and I can help review or test the updated version if useful.

morgan-coded:
Pushed the rename — backend key, factory, docs, and tests are all `milvus` now (from `milvuslite`), per your note. This can be the single Milvus PR; @zc277584121 can close #1129 as superseded whenever.

davidmezzetti:
Thanks. I see #1129 was closed so I'll review this one. At first glance it looks basically the same as the `zvec` one.

I'll try to get cycles to review all these outstanding PRs but I've been tied up on another project.

PR #1151, opened 2026-07-17
Update Authorization to use standard FastAPI HTTPBearer

### LINKED ISSUE #1150: The authorization header doesn't work on Swagger UI
When I run `txtai` with `TOKEN` envvar, the Swaager UI page add an `authorization` header input field in every API endpoints.
But when I put the token there, it just doesn't works. The generated curl command doesn't add header to the request and end up with 401 error.

Image

link89:
This PR fix the issue #1150

With this fix, the SwaggerUI will provide an Authorz button when `TOKEN` envvar is set. And the auth header will be set correctly.

image

### DIFF
--- a/src/python/txtai/api/authorization.py
+++ b/src/python/txtai/api/authorization.py
@@ -3,9 +3,10 @@ Authorization module
 """
 
 import hashlib
-import os
+import hmac
 
-from fastapi import Header, HTTPException
+from fastapi import Depends, HTTPException
+from fastapi.security import HTTPAuthorizationCredentials, HTTPBearer
 
 
 class Authorization:
@@ -13,17 +14,17 @@ class Authorization:
     Basic token authorization.
     """
 
-    def __init__(self, token=None):
+    def __init__(self, token):
         """
         Creates a new Authorization instance.
 
         Args:
-            token: SHA-256 hash of token to check
+            token: SHA-256 hash of the valid authorization token
         """
 
-        self.token = token if token else os.environ.get("TOKEN")
+        self.token = token
 
-    def __call__(self, authorization: str = Header(default=None)):
+    def __call__(self, authorization: HTTPAuthorizationCredentials = Depends(HTTPBearer())):
         """
         Validates authorization header is present and equal to current token.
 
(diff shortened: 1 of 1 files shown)


davidmezzetti:
I'll have to research this issue more. I'm not crazy about making changes to the authorization scheme for a testing UI.

link89:
Understood. I just wanted to clarify that this modification only affects the generated openapi.json and does not change the authentication process. Its main purpose is to help Swagger Codegen generate more accurate client code.

davidmezzetti:
Is there a way to test this?

link89:
If you want to test it manually, you may try install txtai from my fix branch.

```bash
pip install git+https://github.com/link89/txtai.git@fix-openapi-auth-ui
TOKEN=6ca13d52ca70c883e0f0bb101e425a89e8624de51db2d2392593af6a84118090 CONFIG=service.yml uvicorn txtai.api:app
```
Then check the swagger docs page, e.g http://localhost:8000/docs with token `abc123`.

The attachment [openapi.json](https://github.com/user-attachments/files/30248374/openapi.json) is the spec I dump from my app. You may verify it by upload it to https://editor.swagger.io/

davidmezzetti:
Thanks. I need to research this further. I'm not sure if this is the right way to handle this still.

link89:
There is a dedicated page about extending openapi spec in FastAPI document: https://fastapi.tiangolo.com/how-to/extending-openapi/ , just for your reference.

davidmezzetti:
I have no reason to have a non-standard authorization pattern here. The design is expecting `Authorization: Bearer <token>`.

With that, I believe it's much simpler to just modify the authorization class as follows:

```python
"""
Authorization module
"""

import hashlib
import hmac

from fastapi import Depends, HTTPException
from fastapi.security import HTTPAuthorizationCredentials, HTTPBearer


class Authorization:
    """
    Basic token authorization.
    """

    def __init__(self, token):
        """
        Creates a new Authorization instance.

        Args:
            token: SHA-256 hash of the valid authorization token
        """

        self.token = token

    def __call__(self, authorization: HTTPAuthorizationCredentials = Depends(HTTPBearer())):
        """
        Validates authorization header is present and equal to current token.

        Args:
            authorization: authorization header
        """

        # Authorization header is parsed into {scheme="Bearer", credentials="<token>"}
        if not hmac.compare_digest(self.token, self.digest(authorization.credentials)):
            raise HTTPException(status_code=401, detail="Invalid Authorization Token")

    def digest(self, token):
        """
        Computes a SHA-256 hash for input authorization token.

        Args:
            token: authorization token

        Returns:
            SHA-256 hash of authorization token
        """

        return hashlib.sha256(token.encode("utf-8")).hexdigest()
```

I tested that locally and it works fine. If you want to update the PR with that, I'm happy to let you take the credit.

link89:
It's much more clean to use build-in API, I have update the PR with your suggested method and confirm that it does work.

davidmezzetti:
Thank for the PR, I appreciate it!

PR #1159, opened 2026-07-23
Fix HNSW search RuntimeError when limit exceeds indexed element count

### LINKED ISSUE #1158: HNSW search crashes with RuntimeError when limit exceeds indexed element count
The HNSW backend crashes when a search `limit` exceeds the number of indexed elements, instead of returning the available results like the other ANN backends do.

## Reproduction

```python
from txtai.ann import ANNFactory
import numpy as np

data = np.random.rand(3, 240).astype(np.float32)
data /= np.linalg.norm(data, axis=1)[:, np.newaxis]

ann = ANNFactory.create({"backend": "hnsw", "dimensions": 240})
ann.index(data)

ann.search(data[0:1], 10)  # limit (10) > indexed count (3)
```

```
RuntimeError: Cannot return the results in a contiguous 2D array. Probably ef or M is too small
```

Confirmed on current `main` (`a1224c06`).

## Root cause

`HNSW.search` in `src/python/txtai/ann/dense/hnsw.py` passes the raw `limit` straight through as `k` to `hnswlib`'s `knn_query`:

```python
ids, distances = self.backend.knn_query(queries, k=limit)
```

`hnswlib` raises when `k` is larger than the number of elements currently in the index. Every other dense backend handles `limit > count` gracefully: `faiss` pads with `-1`, `annoy` tolerates `n > items`, and the SQL-backed backends apply `LIMIT`. HNSW is the one backend that crashes instead of returning what it has.

This is easy to hit in practice: any query against a small or freshly-seeded index with the default `limit=3` in `Search.__call__`, or any explicit `limit` larger than the current document count, will crash.

AmirF194:
## Root cause

`HNSW.search` (`src/python/txtai/ann/dense/hnsw.py`) passes the raw `limit` straight through as `k` to hnswlib's `knn_query`:

```python
ids, distances = self.backend.knn_query(queries, k=limit)
```

hnswlib raises when `k` exceeds the number of elements currently in the index. Every other dense backend already tolerates `limit > count`: faiss pads with `-1`, annoy tolerates `n > items`, and the SQL-backed backends apply `LIMIT`. HNSW is the only one that crashes.

## Fix

Clamp `k` to `min(limit, self.count())` before calling `knn_query`. `self.count()` is the same accessor the backend already exposes (`get_current_count() - deletes`), so this keeps deletes correctly out of the clamp.

## Test

Added `testHnswSearchLimitExceedsCount`, which indexes 3 vectors and searches with `limit=10`. Fails on main with `RuntimeError: Cannot return the results in a contiguous 2D array. Probably ef or M is too small`, passes on this branch.

## Verified

Docker (`python:3.12-slim`, `pip install -e .` plus `hnswlib`/`numpy`/`sqlalchemy`), differential main vs. this branch, `module.__file__` provenance checked on both sides. Ran the full `testann.testdense` suite on this branch: the three HNSW tests (`testHnsw`, `testHnswCustom`, `testHnswSearchLimitExceedsCount`) pass; the other 37 failures in that run are pre-existing `ModuleNotFoundError`s from backends I did not install locally (annoy, ggml, milvus, torch, fastapi, etc.), unrelated to this change. `black --check` and `pylint` (10.00/10) both pass on the two changed files.

Fixes #1158

### DIFF
--- a/src/python/txtai/ann/dense/hnsw.py
+++ b/src/python/txtai/ann/dense/hnsw.py
@@ -85,8 +85,8 @@ class HNSW(ANN):
         if ef:
             self.backend.set_ef(ef)
 
-        # Run the query
-        ids, distances = self.backend.knn_query(queries, k=limit)
+        # Run the query, clamp k to the current element count as hnswlib errors when k exceeds it
+        ids, distances = self.backend.knn_query(queries, k=min(limit, self.count()))
 
         # Map results to [(id, score)]
         results = []
--- a/test/python/testann/testdense.py
+++ b/test/python/testann/testdense.py
@@ -196,6 +196,21 @@ class TestDense(unittest.TestCase):
         # Test with custom settings
         self.runTests("hnsw", {"hnsw": {"efconstruction": 100, "m": 4, "randomseed": 0, "efsearch": 5}})
 
+    def testHnswSearchLimitExceedsCount(self):
+        """
+        Test Hnswlib backend search when limit exceeds the indexed element count
+        """
+
+        data = np.random.rand(3, 240).astype(np.float32)
+        self.normalize(data)
+
+        ann = ANNFactory.create({"backend": "hnsw", "dimensions": 240})
+        ann.index(data)
+
+        # Query limit (10) exceeds the number of indexed elements (3), previously raised a RuntimeError
+        results = ann.search(data[0:1], 10)
+        self.assertEqual(len(results[0]), 3)
+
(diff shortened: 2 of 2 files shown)


davidmezzetti:
Nice catch! I'll merge this in but I suggest putting a fix upstream too (https://github.com/nmslib/hnswlib)

AmirF194:
Thanks for merging! Good idea on the hnswlib upstream report, I'll look into filing one there since the same unclamped k issue likely affects other callers of knn_query.

PR #1171, opened 2026-08-02
Align LEMUR names with txtai style

### LINKED ISSUE #1169: Update LEMUR to follow `txtai` coding conventions
When you get a moment @morgan-coded please take a look at LEMUR and update to more closely adhere to the coding style in the rest of txtai. The project doesn't often use underscore separated variable names is the main one. Appreciate it.

morgan-coded:
LEMUR's trainer and pooling code used underscore-separated names for internal and public settings, which differed from the concatenated lowercase style used across txtai. This renames the fit and trainer arguments, config keys, model attributes, and local variables, including the training-only `outputlayer` state prefix, while leaving behavior unchanged. The docs and existing artifact round-trip tests use the new names. Closes #1169

### DIFF
--- a/docs/pipeline/train/lemur.md
+++ b/docs/pipeline/train/lemur.md
@@ -23,24 +23,24 @@ trainer(
     corpus,
     "lemur-model",
     epochs=100,
-    validation_split=0.1,
+    validationsplit=0.1,
 )
 ```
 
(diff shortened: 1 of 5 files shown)


davidmezzetti:
Thank you!

Issue #1170, opened 2026-08-02
LEMUR Improvements


davidmezzetti:
@morgan-coded 

Couple thoughts after using the LEMUR pipeline for some testing.

- It feels like the training code should all be in the training pipeline (right now it's split between the pooling method and training pipeline)
- It would be great if the LEMUR training pipeline had some feedback (i.e. tqdm and up to date validation loss and % done) like the HF Trainer pipeline
- I'm wondering if we could create a default LEMUR model that could be loaded (something like `lemur: True`). I think I see ELM is one attempt at that. I wonder we just trained a large scale model using a dataset like this: https://huggingface.co/datasets/lightonai/embeddings-fine-tuning

Overall this is really great work. Multi-vector models are hot right now and this could end up getting a lot of attention.

morgan-coded:
Agreed on consolidating the training path. Once #1171 lands, I'd move training-data construction, feature-model setup, the epoch loop, and validation selection into `LemurTrainer`, while keeping artifact loading and query-feature/document-weight encoding in the pooler. I'd fold `tqdm` progress with live validation loss and percent complete into that move, following the existing HF Trainer path. I can also take the `lemur: True` load path, config plumbing, and tests, but I'd leave the default artifact, training run, dataset choice, and hosting call to you. The LightOn set's query/document/score split, with a positive followed by mined negative candidates, gives that experiment a usable starting shape without locking the choice in.

davidmezzetti:
All sounds good. 

Let's table the default `lemur` model for now. At some point later I'll take a look at building a default model.

PR #1172, opened 2026-08-03
Consolidate LEMUR training in the trainer pipeline

morgan-coded:
Moves LEMUR's training path — training-data construction, feature-model setup, the epoch loop, and validation selection — from the pooling module into `LemurTrainer`, following the plan approved in #1170. Artifact loading, query-feature/document-weight encoding, and save stay in the pooler, while the nn modules remain shared by both sides. Training now reports percent complete and live validation loss through `tqdm`, while non-interactive runs stay quiet and minimal installs use an iterable fallback when `tqdm` is unavailable. With a fixed seed, artifacts before and after the move are byte-identical (config and all seven tensors), and tests moved with their code plus a new progress test. `lemur: True` load-path sugar stays out of this change, as tabled in #1170.


davidmezzetti:
Thanks for this. Given that `tqdm` is a dependency of `transformers` perhaps it makes sense to add to the default library requirements section in setup.py like the others (i.e. safetensors etc). Then there would be a library.tqdm import?

davidmezzetti:
> Thanks for this. Given that `tqdm` is a dependency of `transformers` perhaps it makes sense to add to the default library requirements section in setup.py like the others (i.e. safetensors etc). Then there would be a library.tqdm import?

Doing this will also help with the missing test coverage: https://coveralls.io/jobs/185594771/source_files/8818230347

morgan-coded:
Done — `tqdm>=4.66.3` is now in the default dependencies and the trainer imports it through `library.tqdm()` following the safetensors pattern, so the inline fallback and its uncovered branch are gone; minimal installs still import cleanly through the library stub. Tests, Black and Pylint pass locally on the new head.

morgan-coded:
Flagging an ecosystem break before it reaches master: `testDocling` now fails deterministically on windows-latest with `ConversionError ... InvalidCxxCompiler: Compiler: cl is not found`, raised from torchdynamo inside the docling conversion path. It reproduced identically on two runs at this PR's head on my fork, while the same job was green a day earlier at the previous head. The installed-package diff between the green and failing runs shows exactly seven version drifts, headlined by docling 2.117.0 -> 2.118.0 with docling-core 2.89.0 -> 2.90.0 — the library in the traceback; nothing in this PR touches that path, and the ubuntu leg is green. Master's last windows run predates the docling release, so the first push after any merge would surface this and look like the merge caused it. Pinning `docling<2.118` or reporting the compile-path regression upstream both work; can take either.

davidmezzetti:
I'll merge this in the AM.

I still think it's going to be missing some test coverage as there is no path I see for testing when tqdm is not installed. That would go in testlibrary.py.

morgan-coded:
Added the tqdm stub path to testlibrary.py — the hidden-module list now includes `tqdm`/`tqdm.auto`, and the suite asserts `lib.tqdm().tqdm` raises the install-hint `ImportError` alongside the other core libraries. Passes locally with Black clean and no new Pylint messages; nothing else changed.

davidmezzetti:
Thanks - all looks good! I appreciate all the diligent work on this.

I'll merge and handle the Docling issue separately.

```
docling.exceptions.ConversionError: Conversion failed for: article.pdf with status: failure. Errors: InvalidCxxCompiler: Compiler: cl is not found.
```

The build has a C compiler available so I'm not sure what is happening.

PR #1175, opened 2026-08-04
fix: create default agent tools on first use

### LINKED ISSUE #1174: Agent disabled and smolagents wrongly blamed when installing only the "agent" extra
Installing txtai with the documented `agent` extra disables agents entirely, and the error blames `smolagents` — which is installed and importable.

```
pip install "txtai[agent]"
```

```python
from txtai.agent import Agent

Agent(llm="hf-internal-testing/tiny-random-LlamaForCausalLM", max_steps=1)
```

```
ImportError: smolagents is not available - install "agent" extra to enable
```

The `agent` extra *is* installed, and `import smolagents` works fine (1.26.0 here). `txtai.Agent` has silently become the placeholder stub:

```python
>>> from txtai import Agent
>>> Agent.__module__
'txtai.agent.placeholder'
```

## Cause

`ToolFactory.DEFAULTS` builds a tool instance for every default alias in the class body, so all ten are constructed at import time:

```python
    # Default toolkit
    DEFAULTS = {
        "bash": BashTool(),
        ...
        "read": ReadTool(),
        ...
    }
```

`ReadTool.__init__` creates a `Textractor`, which creates an `HTMLToMarkdown`, which needs `beautifulsoup4`. That ships in `extras["pipeline-data"]`, not `extras["agent"]`, so it raises:

```
ImportError: HTMLToMarkdown pipeline is not available - install "pipeline" extra to enable
```

Because that happens *during import* of `txtai.agent.tool`, it propagates up to `agent/__init__.py`:

```python
# Conditional import
try:
    from .base import Agent
    ...
    from .tool import *
except ImportError:
    from .placeholder import Agent
```

The `except ImportError` catches it and installs the placeholder, whose message names `smolagents` unconditionally. So a missing `pipeline` dependency is reported as a missing `agent` dependency, and every agent feature is disabled — including the nine default tools that have no such dependency and the many configurations that never touch `read` at all.

Measured per-tool with only the `agent` extra installed, loading each tool module directly:

```
BashTool        OK          0.4 ms
EditTool        OK          0.3 ms
GlobTool        OK          1.2 ms
GrepTool        OK          0.4 ms
ReadTool        ImportError  HTMLToMarkdown pipeline is not available - install "pipeline" extra...
TodoWriteTool   OK          0.4 ms
WriteTool       OK          0.3 ms
PythonInterpreterTool OK    0.1 ms
UserInputTool   OK          0.0 ms
WebSearchTool   OK          0.0 ms
```

Nine of ten construct without the `pipeline` extra. Only `read` needs it, and only when it is actually requested.

## Expected behavior

Installing `txtai[agent]` should give a working agent. A default tool whose optional dependency is missing should fail when that tool is requested, with an error naming the extra that is actually missing.

## Why CI doesn't catch it

`build.yml` installs `.[all,dev]`, so `beautifulsoup4` is always present and `DEFAULTS` always builds. `minimal.yml` only runs `import txtai`, which succeeds because the failure is swallowed into the placeholder. Nothing exercises the `agent` extra on its own.

## Environment

- txtai 9.13.0 (`main`, 8383e6c)
- Python 3.12.10, Windows
- Installed: `jinja2`, `mcp<2.0`, `mcpadapt`, `smolagents` 1.26.0, `transformers`, `torch` — i.e. `extras["agent"]` plus the base install, no `pipeline` extra

LHMQ878:
## Summary

Fixes #1174.

`pip install "txtai[agent]"` disabled agents entirely, and the error blamed `smolagents` — which was installed and importable. This makes the default tools build on first request instead of at import time, so a tool with a missing optional dependency only fails when that tool is actually asked for, with an error naming the extra that is really missing.

## Root cause

`ToolFactory.DEFAULTS` built an instance for every alias in the class body, so all ten default tools were constructed while `txtai.agent.tool` was being imported:

```python
    # Default toolkit
    DEFAULTS = {
        "bash": BashTool(),
        ...
        "read": ReadTool(),
        ...
    }
```

`ReadTool.__init__` creates a `Textractor`, which creates an `HTMLToMarkdown`, which needs `beautifulsoup4`. That ships in `extras["pipeline-data"]`, not `extras["agent"]`, so with only the documented `agent` extra installed the class body raised:

```
ImportError: HTMLToMarkdown pipeline is not available - install "pipeline" extra to enable
```

Because that happens during import, it reaches the conditional import in `agent/__init__.py`:

```python
# Conditional import
try:
    from .base import Agent
    ...
    from .tool import *
except ImportError:
    from .placeholder import Agent
```

`except ImportError` swallows it and installs the placeholder, whose message names `smolagents` unconditionally. Measured symptom — no exception on `import txtai`, just a silently substituted stub:

```python
>>> from txtai import Agent
>>> Agent.__module__
'txtai.agent.placeholder'
```

Nine of the ten defaults have no such dependency. Measured with only the `agent` extra installed, loading each tool module directly:

| Tool | Result | Time |
|:--|:--|--:|
| BashTool | OK | 0.4 ms |
| EditTool | OK | 0.3 ms |
| GlobTool | OK | 1.2 ms |
| GrepTool | OK | 0.4 ms |
| ReadTool | `ImportError` — install "pipeline" extra | — |
| TodoWriteTool | OK | 0.4 ms |
| WriteTool | OK | 0.3 ms |
| PythonInterpreterTool | OK | 0.1 ms |
| UserInputTool | OK | 0.0 ms |
| WebSearchTool | OK | 0.0 ms |

So one tool's optional dependency disabled all agent functionality, including configurations that never reference `read`.

## Fix

`DEFAULTS` now holds constructors, and `ToolFactory.default(name)` builds each tool on first request:

```python
    # Cache of default tool instances, keyed by constructor
    INSTANCES = {}

    @staticmethod
    def default(name):
        constructor = ToolFactory.DEFAULTS[name]
        if constructor not in ToolFactory.INSTANCES:
            ToolFactory.INSTANCES[constructor] = constructor()

        return ToolFactory.INSTANCES[constructor]
```

Caching is keyed by constructor rather than by alias, so `webview` and `read` still resolve to one shared instance the way `DEFAULTS["webview"] = DEFAULTS["read"]` did before. Both `create()` call sites go through `default()`.

## Behavior parity

`tools=["defaults"]` output is byte-identical to the unmodified tree:

```
defaults count: 10
bash, edit, glob, grep, python, question, read, todowrite, websearch, write
unique instances: 10
read is webview: True
```

With only the `agent` extra installed, `Agent.__module__` is now `txtai.agent.base` and agents construct normally. Requesting `read` without the `pipeline` extra raises `ImportError: HTMLToMarkdown pipeline is not available - install "pipeline" extra to enable` — naming the extra that is actually missing, at the point of use.

## Tests

Adds `TestToolFactory` to `test/python/testagent.py` — 11 tests covering that `DEFAULTS` holds constructors, that `default()` resolves and caches, that aliases share one instance, deferred construction (a counting stub), a broken tool raising only when requested, both `create()` paths, and the import-time regression.

The import-time test is the one that pins the reported bug. It spawns a subprocess with `sys.modules["bs4"] = None` to stand in for an install without the `pipeline` extra, then asserts `Agent.__module__`:

```
AssertionError: 'txtai.agent.placeholder' != 'txtai.agent.base'
```

These live in `testagent.py` rather than a new `testtoolfactory.py` on purpose: the `coverage` target in the `Makefile` enumerates explicit `-k` module prefixes, so a new file name matching none of them would never be collected in CI. I verified this — under the existing `-k` list a standalone file collected 0 of the new tests, while as a class in `testagent.py` all 11 are collected.

Red check in their final location, against the unmodified source:

```
9 failed, 11 passed      # 8 of the 11 new tests, plus 1 pre-existing failure (below)
```

With the fix:

```
1 failed, 19 passed      # the same pre-existing failure
```

The three that pass either way are `testDefaultCachesAliases`, `testCreateEmpty` and `testCreateDefaultAlias` (`create()` already worked for a plain alias); they guard the surrounding contract rather than the regression.

## Verification

- pylint `10.00/10`. I confirmed the baseline on the unmodified tree is also `10.00/10`, so this doesn't ride on a pre-existing deduction.
- black clean.
- `docs/agent/configuration.md`: additive 3-line note that `read`/`webview` needs the `pipeline` extra.

Two disclosures:

- The one remaining failure, `testToolsDefaults`, fails identically before and after this change on my machine. It is not related: `bash` runs `subprocess(..., text=True)`, which hits `UnicodeDecodeError: 'gbk' codec can't decode byte 0xbf` under my Chinese Windows locale. Confirmed by running it on a pristine checkout.
- I could not run the full `.[all,dev]` suite locally. I ran `testagent.py` (with the `agent` extra plus `beautifulsoup4`, `nltk`, `pandas`) and the standalone-extra probes described above.

### DIFF
--- a/docs/agent/configuration.md
+++ b/docs/agent/configuration.md
@@ -99,9 +99,6 @@ The following shortcut strings load tools directly. Passing a Tool instance is a
 | write       | Writes content to file                                    |
 | *.md        | Loads a [`skill.md`](https://agentskills.io/specification) file |
 
-Tools are created when requested. The `read` tool (and its `webview` alias) extracts text with the
-[Textractor](../../pipeline/data/textractor) pipeline, so it additionally needs the `pipeline` extra.
-Requesting it without that extra installed raises an error naming the missing extra.
 
 ## instructions
 
--- a/docs/agent/configuration.md
+++ b/docs/agent/configuration.md
@@ -99,6 +99,10 @@ The following shortcut strings load tools directly. Passing a Tool instance is a
 | write       | Writes content to file                                    |
 | *.md        | Loads a [`skill.md`](https://agentskills.io/specification) file |
 
+Tools are created when requested. The `read` tool (and its `webview` alias) extracts text with the
+[Textractor](../../pipeline/data/textractor) pipeline, so it additionally needs the `pipeline` extra.
+Requesting it without that extra installed raises an error naming the missing extra.
+
 ## instructions
 
 ```yaml
--- a/setup.py
+++ b/setup.py
@@ -52,7 +52,7 @@ extras["dev"] = [
     "pylint",
 ]
 
-extras["agent"] = ["jinja2>=3.1.6", "mcp<2.0", "mcpadapt>=0.1.0", "smolagents>=1.23"]
+extras["agent"] = ["beautifulsoup4>=4.9.3", "jinja2>=3.1.6", "mcp<2.0", "mcpadapt>=0.1.0", "smolagents>=1.23"]
 
 extras["ann"] = [
     "annoy>=1.16.3",
--- a/src/python/txtai/agent/tool/factory.py
+++ b/src/python/txtai/agent/tool/factory.py
@@ -30,22 +30,53 @@ class ToolFactory:
     Methods to create tools.
     """
 
-    # Default toolkit
+    # Default toolkit. Stores constructors, not instances - these are built on first use by `default`.
+    #
+    # Instances must not be created here. A class body runs at import time, so a tool that needs an
+    # optional dependency would raise while `txtai.agent` is being imported. That ImportError is
+    # caught in agent/__init__.py, which then installs the placeholder Agent and reports that
+    # smolagents is missing - regardless of which dependency was actually missing. ReadTool builds a
+    # Textractor, which needs the "pipeline" extra, so installing only the documented "agent" extra
+    # disabled agents entirely and pointed at the wrong package.
     DEFAULTS = {
-        "bash": BashTool(),
-        "edit": EditTool(),
-        "glob": GlobTool(),
-        "grep": GrepTool(),
-        "python": PythonInterpreterTool(),
-        "question": UserInputTool(),
-        "read": ReadTool(),
-        "todowrite": TodoWriteTool(),
-        "websearch": WebSearchTool(),
-        "write": WriteTool(),
+        "bash": BashTool,
+        "edit": EditTool,
+        "glob": GlobTool,
+        "grep": GrepTool,
+        "python": PythonInterpreterTool,
+        "question": UserInputTool,
+        "read": ReadTool,
+        "todowrite": TodoWriteTool,
+        "websearch": WebSearchTool,
+        "write": WriteTool,
     }
 
     # Backwards compatible mappings
     DEFAULTS["webview"] = DEFAULTS["read"]
+
+    # Cache of default tool instances, keyed by constructor
+    INSTANCES = {}
+
+    @staticmethod
+    def default(name):
+        """
+        Gets a default tool by alias name, creating it on first use.
+
+        Instances are cached, so an alias and its backwards compatible mapping share a single tool,
+        as do repeated agents in the same process.
+
+        Args:
+            name: default tool alias name
+
+        Returns:
+            Tool
+        """
+
+        constructor = ToolFactory.DEFAULTS[name]
+        if constructor not in ToolFactory.INSTANCES:
+            ToolFactory.INSTANCES[constructor] = constructor()
+
+        return ToolFactory.INSTANCES[constructor]
 
     @staticmethod
     def create(config):
@@ -81,11 +112,11 @@ class ToolFactory:
 
             # Get default tool, if applicable
             elif isinstance(tool, str) and tool in ToolFactory.DEFAULTS:
-                tool = ToolFactory.DEFAULTS[tool]
+                tool = ToolFactory.default(tool)
 
             # Get ALL default tools, if applicable
             elif isinstance(tool, str) and tool == "defaults":
-                tools.extend(set(ToolFactory.DEFAULTS.values()))
+                tools.extend({ToolFactory.default(name) for name in ToolFactory.DEFAULTS})
                 tool = None
 
             # Support importing MCP tool collections
--- a/src/python/txtai/agent/tool/factory.py
+++ b/src/python/txtai/agent/tool/factory.py
@@ -30,40 +30,25 @@ class ToolFactory:
     Methods to create tools.
     """
 
-    # Default toolkit. Stores constructors, not instances - these are built on first use by `default`.
-    #
-    # Instances must not be created here. A class body runs at import time, so a tool that needs an
-    # optional dependency would raise while `txtai.agent` is being imported. That ImportError is
-    # caught in agent/__init__.py, which then installs the placeholder Agent and reports that
-    # smolagents is missing - regardless of which dependency was actually missing. ReadTool builds a
-    # Textractor, which needs the "pipeline" extra, so installing only the documented "agent" extra
-    # disabled agents entirely and pointed at the wrong package.
-    DEFAULTS = {
-        "bash": BashTool,
-        "edit": EditTool,
-        "glob": GlobTool,
-        "grep": GrepTool,
-        "python": PythonInterpreterTool,
-        "question": UserInputTool,
-        "read": ReadTool,
-        "todowrite": TodoWriteTool,
-        "websearch": WebSearchTool,
-        "write": WriteTool,
-    }
-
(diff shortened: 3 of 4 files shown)


davidmezzetti:
Hello, interesting find! 

I think it would be much more straightforward to just add `beautifulsoup4>=4.9.3` to the optional agent dependencies. I consider the `ReadTool` pretty critical path and bs4 isn't that big a library (~100K).

LHMQ878:
Thanks — I tried your suggestion in a clean venv before replying, and I think it fixes the import but leaves `ReadTool` quietly broken. Both changes look worth having, for different reasons.

### Setup

A fresh venv with exactly the documented `agent` extra (`jinja2`, `mcp`, `mcpadapt`, `smolagents`) plus txtai's base deps — no `pipeline` extra:

```
                                        agents usable?   ReadTool('report.docx')
A  base + bs4 only  (your suggestion)   yes              mojibake, no error
B  this PR, no bs4                      yes              ImportError naming "pipeline"
C  this PR + bs4                        yes              mojibake, no error
D  base, no bs4     (status quo)        NO               n/a
```

Row D is the reported bug, and it reproduced exactly — `smolagents` **is** installed, and the error still says `smolagents is not available - install "agent" extra to enable`, because the `ImportError` from the class body is swallowed by the `try` in `agent/__init__.py`.

### Why bs4 alone isn't quite enough

`bs4` unblocks `HTMLToMarkdown`, which is the only unconditional raise in the `Textractor` chain — so you're right that it's the sole *construction* blocker today. But `FileToHTML` doesn't raise; it degrades:

```python
# filetohtml.py
backend = "tika" if Tika.available() else "docling" if Docling.available() else "liteparse" if LiteParse.available() else None
self.backend = ... else None
```

and `Textractor.__call__` then takes the no-backend branch:

```python
elif self.html:      # None without tika/docling/liteparse
    ...
else:
    html = self.safeopen.retrieve(path)   # raw bytes of the .docx
return self.markdown(html)
```

Measured on row A (bs4 installed, no `pipeline`):

```
ReadTool constructed OK
  FileToHTML backend : None
  read('a.txt')  -> 'hello world from a text file'          (fine)
  read('b.docx') -> '偋̄〰〰〰〰〰〰〰〰〰〰〰〰〰〰〰〰...'   (raw bytes, no error)
```

So with bs4 alone the agent gets a tool whose stated job is "automatically extracts text content from binary files", and on a PDF/DOCX it silently returns garbage that the LLM will treat as document content. That failure is harder to diagnose than an ImportError, and it's the one an agent can't recover from.

### What I'd suggest

**Both**, and they don't conflict:

- Add `beautifulsoup4>=4.9.3` to the `agent` extra, as you propose — it makes `read` work for text and HTML out of the box, which matches treating it as critical path.
- Keep the lazy construction from this PR, because it's what stops *any* future optional dependency in a default tool from disabling agents wholesale and misattributing the cause. Right now `read` is the only default tool that reaches into `pipeline` (I checked all ten), so bs4 covers today's case exactly; the class-body construction is the reason a single tool's dependency can take down the import at all.

Happy to add the `beautifulsoup4` line to `extras["agent"]` in this PR so it's one change. If you'd rather drop the factory refactor entirely and just ship the dependency, that's your call and I'll trim it down — but in that case it may be worth having `FileToHTML` signal the no-backend case rather than falling through to raw bytes, since that path stays reachable for anyone installing `agent` without `pipeline`.

LHMQ878:
Separately on CI: both workflow runs on this branch are sitting at `action_required` (they need a maintainer to approve the run — the `feat/lemur-trainer` branch is in the same state, so it isn't specific to this PR). I ran both jobs locally in the meantime:

```
minimal   MINIMAL=1 pip install .  ->  ok
          python -c "import txtai"  ->  ok

build     pytest test/python/testagent.py -k ToolFactory  ->  11 passed
```

So the `unstable` state is the pending approval, not a failure.

davidmezzetti:
I would argue that any default tool should have it's deps as part of the optional dependency. This is an oversight and a good catch.

But I'm not opposed to creating the default tools lazily. I just think the tests being added are way too verbose and too much. I'd have one single new test method to test that. And instead of having the constructors like that I'd just move to a static method that creates the tools on the fly. There is no reason they need to be singletons.

LHMQ878:
Thanks, I’ve updated this based on your feedback. Default tools are now created lazily by a static method on each request, with no cached singleton instances. I also reduced the coverage to one focused test method and added beautifulsoup4>=4.9.3 to the agent extra so ReadTool’s dependency is included.

davidmezzetti:
Sorry for the delay. Can you make this a map lookup for the if/else branch logic? The map defaults can be defined in the method itself vs a global var.

LHMQ878:
Updated — `ToolFactory.default` now uses an in-method map lookup instead of the if/else chain, with the map defined inside the method rather than as a global.

davidmezzetti:
LGTM - I appreciate the PR!

PR #1178, opened 2026-08-09
Convert MUVERA algorithm from NumPy to Torch

### LINKED ISSUE #1167: Convert MUVERA algorithm from NumPy to Torch
The current MUVERA algorithm in `txtai` is written using NumPy. Consider using Torch to improve performance and speed.

Aryan-Pardeshi:
Closes #1167

Ports the MUVERA encoding to Torch. Each NumPy operation in `__call__` has a direct counterpart: `np.vstack` becomes `torch.cat`, `np.repeat` becomes `torch.repeat_interleave`, and the two `np.add.at` scatter-accumulates become `Tensor.index_add_`. `__call__` still returns a NumPy array, so nothing downstream changes.

The projection matrices in `random()` and `reducer()` deliberately keep NumPy's seeded generator and are moved across with `torch.from_numpy`. Torch and NumPy RNGs do not agree for the same seed, so generating them with Torch would have silently changed every encoding produced by a given seed. Keeping NumPy there makes this port bit-identical rather than merely equivalent.

I verified that by generating a reference encoding from the current NumPy implementation on `master` and comparing against the Torch output:

```
master-numpy shape (3, 256) | torch shape (3, 256)
max abs diff: 0.0
```

That reference is committed as `test/muvera_baseline.npy` and asserted by `testMuveraTorchMatchesBaseline`, so a future change that alters encodings for a fixed seed will fail the suite rather than pass silently. `testMuveraTorchMatchesNumPy` covers output width and determinism across repeated calls.

Checks run locally: `pytest test/python/testmodels/testpooling.py` passes, pylint with the flags from `.pre-commit-config.yaml` rates both touched files 10.00/10, and black reports them unchanged.

One decision worth your call: I kept the NumPy dependency for the RNG rather than removing it. If you would rather this be fully Torch-native and accept that encodings change for a given seed, say so and I will regenerate the baseline.


davidmezzetti:
Hello - Awesome work! Can you just remove the unit test? I don't want to store npy files in GitHub as it could be perceived as a security risk, thank you.

Aryan-Pardeshi:
Done — removed `testMuveraTorchMatchesBaseline` and the committed `muvera_baseline.npy` fixture. `testMuveraTorchMatchesNumPy` stays; it covers output shape and determinism without needing any stored file, so the equivalence check isn't lost, just the part that required checking in binary data.

davidmezzetti:
Thanks, I'll run the tests and merge assuming no issues.

Aryan-Pardeshi:
Happy to contribute 😄

PR #1183, opened 2026-08-11
Fix OpenAI chat message selection

### LINKED ISSUE #1182: OpenAI chat route sends the oldest message to non-LLM backends
### Describe the bug

The OpenAI-compatible `/v1/chat/completions` route always selects `messages[0]["content"]` before dispatching requests to a named agent, the `embeddings` search model, a non-LLM pipeline, or a workflow.

OpenAI chat messages are ordered oldest to newest. A normal request that starts with a system message therefore sends the system prompt, not the current user prompt, to these txtai backends. Multi-turn requests likewise replay the oldest message.

The default `llm` path is not affected because it forwards the complete `messages` list.

### Reproduction

```python
messages = [
    {"role": "system", "content": "You are concise"},
    {"role": "user", "content": "What is retrieval augmented generation?"},
]
```

Dispatching this request to a configured non-LLM pipeline currently calls the pipeline with:

```text
You are concise
```

instead of:

```text
What is retrieval augmented generation?
```

The same behavior is visible for named agents, `model="embeddings"`, and workflows because all four paths share the value assigned from `messages[0]`.

### Expected behavior

Non-LLM chat backends should process the latest message in the request, while the default LLM path should continue receiving the full conversation.

### Proposed fix

Select the final message for the single-message backends and add a regression test containing both a system message and a user message.

FU-max-boop:
## Summary

- use the latest chat message when dispatching OpenAI-compatible requests to agents, embeddings search, non-LLM pipelines, and workflows
- keep the default LLM path unchanged so it still receives the full conversation
- add a regression test with a system message followed by the current user message

Fixes #1182

## Validation

- `black --check src/python/txtai/api/routers/openai.py test/python/testapi/testopenai.py`
- `pylint --rcfile=.pylintrc src/python/txtai/api/routers/openai.py test/python/testapi/testopenai.py` (10.00/10)
- `python -m py_compile src/python/txtai/api/routers/openai.py test/python/testapi/testopenai.py`
- isolated dispatch matrix covering agent, embeddings, pipeline, workflow, and unchanged full-message LLM behavior

### DIFF
--- a/src/python/txtai/api/routers/openai.py
+++ b/src/python/txtai/api/routers/openai.py
@@ -44,8 +44,8 @@ def chat(
     # Build keyword arguments
     kwargs = {key: value for key, value in [("stream", stream), ("maxlength", max_completion_tokens)] if value}
 
-    # Get first message
-    message = messages[0]["content"]
+    # Get latest message
+    message = messages[-1]["content"]
 
     # Agent
     if model in application.get().agents:
--- a/test/python/testapi/testopenai.py
+++ b/test/python/testapi/testopenai.py
@@ -122,6 +122,24 @@ class TestOpenAI(unittest.TestCase):
 
         self.assertIsNotNone(response["choices"][0]["message"]["content"])
 
+    def testChatLatestMessage(self):
+        """
+        Test a chat completion with multiple messages
+        """
+
+        response = self.client.post(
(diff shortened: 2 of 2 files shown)


davidmezzetti:
LGTM - Interesting catch, thank you!

PR #1184, opened 2026-08-11
docs: add OrcaRouter to the LLM API example notebook

XiaoHuo888-hue:
## Summary

Adds a named [OrcaRouter](https://www.orcarouter.ai) entry to the "Getting started with LLM APIs" example notebook. OrcaRouter is an OpenAI-compatible model routing gateway that exposes 150+ models from OpenAI, Anthropic, Google, DeepSeek, Gemini and more behind a single endpoint and API key. It also provides gateway-level security controls for AI agents.

I'm an engineer on the OrcaRouter team.

## What this changes

- `examples/70_Getting_started_with_LLM_APIs.ipynb`: adds an `ORCAROUTER` entry to `LLM_MODEL_CONFIG` (base URL `https://api.orcarouter.ai/v1`, auth via `ORCAROUTER_API_KEY`, default model `openai/openai/gpt-4o-mini`), a short "OrcaRouter" explainer cell, and threads `api_base`/`api_key` through the notebook's `runllm` helper so the entry runs end to end. Existing providers are unaffected - `api_base` is only forwarded when set.

## Why a named entry rather than the OpenAI-compatible escape hatch

txtai's LiteLLM backend already accepts any OpenAI-compatible endpoint through `api_base`. A named config entry makes [OrcaRouter](https://www.orcarouter.ai) directly discoverable and selectable in the notebook's provider table, alongside Gemini, OpenAI, Claude, Groq and the rest. Named routers such as `openai/orcarouter/auto` pick an upstream model per request, so the same entry can serve the whole model catalog from one key.

## How to use it

1. Get an API key at https://www.orcarouter.ai/console (keys start with `sk-orca-`).
2. Set `ORCAROUTER_API_KEY`.
3. In the notebook, set `IS_ENABLED: True` for `ORCAROUTER` and run; the LLM pipeline routes `openai/gpt-4o-mini` through OrcaRouter. `openai/orcarouter/auto` auto-routes to the best available model per request.

## Verification

- Notebook JSON parses and all code cells compile.
- Live API: ran the notebook's exact `runllm` flow with a real `ORCAROUTER_API_KEY` against `https://api.orcarouter.ai/v1`; `openai/openai/gpt-4o-mini` returned a completion (HTTP 200). Also verified the auto-routing alias `openai/orcarouter/auto` responds.

### DIFF
--- a/examples/70_Getting_started_with_LLM_APIs.ipynb
+++ b/examples/70_Getting_started_with_LLM_APIs.ipynb
@@ -197,6 +197,14 @@
         "    #\n",
         "    # https://docs.voyageai.com/docs/embeddings\n",
         "    'TEXT_EMBEDDING_PATH': 'voyage/voyage-01',\n",
+        "  },\n",
+        "  'ORCAROUTER': {\n",
+        "    'IS_ENABLED': False,\n",
+        "    'ENV_VAR': ['ORCAROUTER_API_KEY'],\n",
+        "    # https://www.orcarouter.ai\n",
+        "    # OpenAI-compatible multi-model routing gateway - one key, one endpoint\n",
+        "    'LLM_MODEL_NAME': 'openai/openai/gpt-4o-mini',\n",
+        "    'API_BASE': 'https://api.orcarouter.ai/v1',\n",
         "  }\n",
         "}\n",
         "\n",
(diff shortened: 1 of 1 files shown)


davidmezzetti:
Hello - Thank you for sharing. I'm not sure updating that notebook and all the related articles makes the most sense. I recommend that you get documentation or Orca on something like LiteLLM's page.

PR #1186, opened 2026-08-15
Add max pooling

### LINKED ISSUE #1185: Add max pooling
Add a new pooling type for max pooling. Should be a simple component like mean pooling.

Aryan-Pardeshi:
Closes #1185

Adds `MaxPooling` alongside the existing mean, cls, last and late pooling components.

**`max.py`** mirrors `MeanPooling`: run the transformers model, then reduce the token embeddings over the sequence dimension. Padded positions are filled with the dtype minimum before the max so they can never win, which keeps a short text encoding identical whether it is encoded alone or batched with a longer one.

**`PoolingFactory`** resolves the new `maxpooling` method, and derives it from the sentence transformers `pooling_mode_max_tokens` config flag the same way cls and last token pooling are derived.

**`HFVectors.ismethod`** accepts `maxpooling` so the method can also be selected through embeddings configuration, and `docs/embeddings/configuration/vectors.md` lists it with the others.

### Tests

`testMax` covers the component: factory resolution, encoding shape, and the padding-independence property. `testMaxMethod` covers deriving the method from the pooling config.

### DIFF
--- a/docs/embeddings/configuration/vectors.md
+++ b/docs/embeddings/configuration/vectors.md
@@ -25,8 +25,8 @@ Embeddings method to use. If the method is not provided, it is inferred using th
 Builds embeddings using a transformers model. While this can be any transformers model, it works best with
 [models trained](https://huggingface.co/models?pipeline_tag=sentence-similarity) to build embeddings.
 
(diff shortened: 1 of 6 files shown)


davidmezzetti:
Thank you for this! The build errors are unrelated so I'll merge this in and fix that in the main branch after.

PR #1188, opened 2026-08-15
Read pooling_mode_mean_tokens with get() so a partial config does not raise

### LINKED ISSUE #1187: PoolingFactory.method raises KeyError on a pooling config without pooling_mode_mean_tokens
`PoolingFactory.method()` reads `pooling_mode_mean_tokens` with a bracket lookup while every other flag on the same lines uses `.get()`:

```python
if config and config.get("pooling_mode_cls_token") and not config["pooling_mode_mean_tokens"]:
    method = "clspooling"

if config and config.get("pooling_mode_lasttoken") and not config["pooling_mode_mean_tokens"]:
    method = "lastpooling"
```

A `1_Pooling/config.json` that sets a pooling mode but omits `pooling_mode_mean_tokens` therefore raises `KeyError` out of `PoolingFactory.create()` rather than resolving a method.

Reproduced on `master` (`3ea1974`):

```python
from unittest.mock import patch
from txtai.models import PoolingFactory

with patch.object(PoolingFactory, "load", return_value={"pooling_mode_cls_token": True}):
    PoolingFactory.method("any/model")
```

```
KeyError: 'pooling_mode_mean_tokens'
```

Same for `{"pooling_mode_lasttoken": True}`.

The surrounding code is defensive everywhere else — `config and config.get(...)` guards each flag, `maxlength()` guards with `if config`, and `load()` swallows `DownloadError` so a missing file returns `None`. This one lookup is the exception, and it turns a partial config into a crash instead of a fallback.

`sentence-transformers` writes the full key set, so stock ST exports are unaffected. Hand-authored pooling configs on the Hub, and configs written by other tooling, are not guaranteed to.

Expected: a missing flag reads as disabled, same as an explicit `false` — `{"pooling_mode_cls_token": true}` resolves to `clspooling`, `{}` resolves to the `meanpooling` default.

I have a fix and test ready and will open a PR referencing this.

Aryan-Pardeshi:
Closes #1187

`PoolingFactory.method()` guarded every other pooling flag with `.get()` but read `pooling_mode_mean_tokens` with a bracket lookup, so a `1_Pooling/config.json` that sets a mode without that key raised `KeyError` out of `PoolingFactory.create()` instead of resolving a method:

```python
with patch.object(PoolingFactory, "load", return_value={"pooling_mode_cls_token": True}):
    PoolingFactory.method("any/model")   # KeyError: 'pooling_mode_mean_tokens'
```

A missing flag now reads as disabled, matching an explicit `false` and the defensive style used by the rest of the module (`config and config.get(...)`, `maxlength()`'s `if config` guard, `load()` swallowing `DownloadError`).

### Tests

`testPartialPoolingConfig` covers the resolution table both ways — cls/last with the mean key absent and present-and-false both resolve to their method, mean-enabled and an empty config resolve to `meanpooling`.

### Note on #1186

#1186 adds a third flag check on the same lines (`pooling_mode_max_tokens`). Whichever of the two lands second I'll bring in line — it's a one-token change either way.

### DIFF
--- a/src/python/txtai/models/pooling/factory.py
+++ b/src/python/txtai/models/pooling/factory.py
@@ -81,11 +81,11 @@ class PoolingFactory:
         config = PoolingFactory.load(path, "1_Pooling/config.json")
 
         # Set to CLS pooling if it's enabled and mean pooling is disabled
-        if config and config.get("pooling_mode_cls_token") and not config["pooling_mode_mean_tokens"]:
+        if config and config.get("pooling_mode_cls_token") and not config.get("pooling_mode_mean_tokens"):
             method = "clspooling"
 
         # Set to last token pooling if it's enabled and mean pooling is disabled
-        if config and config.get("pooling_mode_lasttoken") and not config["pooling_mode_mean_tokens"]:
+        if config and config.get("pooling_mode_lasttoken") and not config.get("pooling_mode_mean_tokens"):
             method = "lastpooling"
 
         # Check for late interaction pooling
--- a/test/python/testmodels/testpooling.py
+++ b/test/python/testmodels/testpooling.py
@@ -5,6 +5,7 @@ Pooling module tests
 import os
 import tempfile
 import unittest
+import unittest.mock
 
 import numpy as np
 import torch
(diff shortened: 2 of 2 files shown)


davidmezzetti:
Going to close this in lieu of #1192 since that is a first-time contributor. Thank you for this!

PR #1194, opened 2026-08-18
Fix LEMUR collection centering

### LINKED ISSUE #1190: LEMUR training/indexing mismatch with batch-scope centering
I found an interaction between #1168's batch-centering behavior and the LEMUR trainer that I didn't account for when these changes landed. Batch-composition dependence itself was intentional and documented in #1168. The narrower issue is that LEMUR training and indexing apply different centering transforms under the same configuration.

The trainer encodes every text with `batch=1` (`pipeline/train/lemur.py` — the loop at `documents = [pooling.encode([text], batch=1, ...)`), so under scope `batch` the centering mean degenerates to each document's own tokens. Indexing encodes `encodebatch` (default 32) documents per model batch and centers across all of them. The same document can therefore receive different centered token embeddings depending on the path and the other documents sharing its encoding batch.

Reproduced at `3ea19746` with `neuml/colbert-bert-tiny`, `center: true`, 64 synthetic documents (script below):

- Token vectors, trainer-vs-index: maximum absolute coordinate difference 0.305; mean per-document maximum absolute coordinate difference 0.164. The `scope: document` control removes the discrepancy under the same padding path (≤ 7e-7), isolating batch-scope centering as the cause in this reproduction. Deterministic across reruns.
- Exact MaxSim over the two encodings: the ordered top-5 list differed for 8/8 synthetic queries. This does not imply relevance quality declined.
- End-to-end with an `epochs=0` artifact trained through `LemurTrainer`: LEMUR feature-vector maximum absolute coordinate difference 0.036; the ordered top-5 list differed for 8/8 queries; the `scope: document` LEMUR control is ≤ 2e-7.

This demonstrates path inconsistency; it does not establish which resulting ranking has better retrieval quality.

Is batch-scoped centering intended to be supported for LEMUR artifacts? If so, what invariant should training, initial indexing, and later updates share? If not, should LEMUR require a batch-invariant scope such as `document`? I held off on proposing a patch because that contract determines which path should change.

<details><summary>Reproduction script</summary>

```python
# Centering-parity reproduction - txtai @ 3ea197468, neuml/colbert-bert-tiny, CPU
import random
import tempfile

import numpy as np
import torch

from txtai.models import Models, PoolingFactory
from txtai.pipeline import LemurTrainer

MODEL = "neuml/colbert-bert-tiny"
BATCH = 32  # txtai encodebatch default
random.seed(42)
VOCAB = (
    "retrieval index vector token model corpus query document ranking search "
    "neural embedding scoring latency memory training feature centering batch "
    "apple river mountain quantum ledger orbit fabric signal harbor lattice"
).split()
DOCS = [" ".join(random.choices(VOCAB, k=random.randint(4, 60))) for _ in range(64)]
QUERIES = [" ".join(random.choices(VOCAB, k=random.randint(3, 8))) for _ in range(8)]


def pool(center, lemur=None):
    # mirrors LemurTrainer's pooling construction
    return PoolingFactory.create(
        {"method": None, "path": MODEL, "device": Models.deviceid(False), "tokenizer": None,
         "maxlength": None, "modelargs": {"center": center, "muvera": None, "lemur": lemur}}
    )


def asnp(x):
    return x.detach().cpu().numpy() if isinstance(x, torch.Tensor) else np.asarray(x)


def trainer_style(p, texts, category="data"):  # batch=1, as pipeline/train/lemur.py
    return [asnp(p.encode([t], batch=1, category=category)[0]) for t in texts]


def index_style(p, texts, category="data"):  # encodebatch default
    return [asnp(v) for v in p.encode(texts, batch=BATCH, category=category)]


def trim(a, b):  # batch encoding pads token rows; real tokens are the leading rows
    return b[: a.shape[0]] if b.ndim == a.ndim and b.shape[0] > a.shape[0] else b


def report(name, A, B):
    diffs = [np.abs(a - trim(a, b)).max() for a, b in zip(A, B)]
    print(f"[{name}] max={max(diffs):.6g} mean_per_doc_max={np.mean(diffs):.6g}")


def maxsim_top5(q, docs):
    return list(np.argsort([float((q @ d.T).max(axis=1).sum()) for d in docs])[::-1][:5])


def fixed_top5(q, docs):
    return list(np.argsort([float(np.dot(np.ravel(q), np.ravel(d))) for d in docs])[::-1][:5])


pb = pool(center=True)
print("resolved center:", pb.center)  # {'scope': 'batch'}
A, B = trainer_style(pb, DOCS), index_style(pb, DOCS)
report("batch-scope trainer-vs-index", A, B)
Q = trainer_style(pb, QUERIES, category="query")
flips = sum(maxsim_top5(q, A) != maxsim_top5(q, [trim(a, b) for a, b in zip(A, B)]) for q in Q)
print(f"top-5 changed for {flips}/{len(Q)} queries")

pd = pool(center={"scope": "document"})
report("document-scope control", trainer_style(pd, DOCS), index_style(pd, DOCS))

out = tempfile.mkdtemp(prefix="lemur-parity-")
LemurTrainer()(MODEL, DOCS, out, gpu=False, vectors={"center": True}, epochs=0,
               trainsubsetsize=64, learnsubsetsize=2000, olssamplesize=512)
pl = pool(center=True, lemur={"path": out})
LA, LB = trainer_style(pl, DOCS), index_style(pl, DOCS)
report("LEMUR fixed vectors batch-scope", LA, LB)
QL = trainer_style(pl, QUERIES, category="query")
flips = sum(fixed_top5(q, LA) != fixed_top5(q, LB) for q in QL)
print(f"LEMUR top-5 changed for {flips}/{len(QL)} queries")

pld = pool(center={"scope": "document"}, lemur={"path": out})
report("LEMUR document-scope control", trainer_style(pld, DOCS), index_style(pld, DOCS))
```

</details>

morgan-coded:
Closes #1190 with the approach from the thread: since training already passes over the corpus, the trainer now computes the collection token mean there and stores it with the artifact.

- With no explicit `center` setting, `LemurTrainer` encodes the corpus uncentered, computes one mean over all data-token rows, centers both data and learn matrices with it (existing `centerdata` collection semantics), runs the full fit on those vectors, and saves the mean as an optional `lemur.center` tensor.
- Loading an artifact that carries the mean auto-selects `collection` centering with it when `center` wasn't explicitly configured. Explicit settings always win; artifacts without the tensor and non-LEMUR paths resolve exactly as before (`batch` stays the default, per the thread).
- Training, indexing, and querying now center identically: the acceptance test encodes the same documents at `batch=1` and `batch=32` through the artifact path and requires identical vectors — it fails on current master (64/64 elements differ) and passes here, in both `epochs=0` and trained modes.

Tests: new integration test (stored mean, both-category parity, explicit-config precedence, old-artifact fallback) plus the full pooling (21/21) and trainer (25/25) modules; pylint and black clean.

### DIFF
--- a/docs/embeddings/configuration/vectors.md
+++ b/docs/embeddings/configuration/vectors.md
@@ -150,9 +150,10 @@ center: boolean | dict
 ```
 
 Centers normalized late-interaction token vectors and normalizes them again before MUVERA or LEMUR encoding. Centering is applied to
-both query and document vectors. `true` uses batch scope and `false` disables centering. When omitted, centering defaults to batch
-scope only for late-interaction models that load more than one `torch.nn.Linear` layer; models with zero or one loaded linear layer keep
-their previous output.
+both query and document vectors. `true` uses batch scope and `false` disables centering. An explicit setting always takes precedence.
+When omitted, a LEMUR artifact with a stored collection mean uses collection scope. Other LEMUR artifacts and late-interaction models
+default to batch scope only when they load more than one `torch.nn.Linear` layer; models with zero or one loaded linear layer keep their
+previous output.
 
 Dictionary configuration supports the following scopes:
 
@@ -162,8 +163,9 @@ center:
 ```
 
 `document` subtracts the mean of each document or query independently. `batch` subtracts the mean of all real token rows in the current
-model batch. `collection` requires a caller-provided, one-dimensional mean vector through either `mean` (an array-like value) or `path`
-(a safetensors file containing a `center.mean` tensor), but not both. Zero-padding rows are never centered and remain zero.
+model batch. Explicit `collection` scope requires a caller-provided, one-dimensional mean vector through either `mean` (an array-like
+value) or `path` (a safetensors file containing a `center.mean` tensor), but not both. LEMUR artifacts can provide this mean
+automatically. Zero-padding rows are never centered and remain zero.
 
 ### muvera
 ```yaml
@@ -184,8 +186,9 @@ lemur:
 Loads a trained [LEMUR](../../../pipeline/train/lemur) fixed dimensional encoder. LEMUR artifacts are corpus-specific and must be
 created with `LemurTrainer` before configuring an embeddings index. Query vectors are summed learned features and document vectors
 are ordinary least squares weights over the artifact's stored token sample. Each artifact contains `config.json` and
-`model.safetensors`; the Safetensors file contains inference state only. See the trainer page for score filtering and Faiss exact/IVF
-search behavior.
+`model.safetensors`; newly trained artifacts also store the collection token mean as `lemur.center`. Loading that mean automatically
+selects collection centering unless `center` is explicitly configured. Artifacts without `lemur.center` retain the previous default.
+The Safetensors file contains inference state only. See the trainer page for score filtering and Faiss exact/IVF search behavior.
 
 ### trust_remote_code
 ```yaml
--- a/docs/pipeline/train/lemur.md
+++ b/docs/pipeline/train/lemur.md
@@ -37,6 +37,10 @@ The learn distribution defaults to `learncategory="query"`. Target documents are
 default encodes selected corpus texts once as data and once as queries. Set `learncategory="data"` to reuse the data encodings, or
 pass a separate iterable of texts with `learn`.
 
+When `vectors.center` is omitted, the trainer encodes the corpus without centering, computes one mean over all corpus token rows and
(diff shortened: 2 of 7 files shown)


davidmezzetti:
LGTM as we discussed! Running the build now and will merge assuming no issues.

davidmezzetti:
Can we please add a test for this condition or remove the validation? https://coveralls.io/jobs/186400239/source_files/9538064581#L206

morgan-coded:
Added — `testLemurCenterMeanValidation` exercises all three rejection branches (2-D, wrong dimension, non-finite). Kept the validation since `fit()` is callable directly with a caller-supplied mean.

Issue #372, opened 2022-10-19
Add Cross-Encoder support to Similarity pipeline


nickchomey:
It seems like Cross Encoders are the preferred model for doing Re-Ranking of search results that were generated by another means (BM25, vector search etc...). However, if I provide one of these models as the path, all the results just have scores of 0.5. 

Sentence Transformers recommends doing this. https://www.sbert.net/examples/applications/retrieve_rerank/README.html

In particular, their msmarco-minilm models seem ideal as a default (maybe the L-6 version?) https://www.sbert.net/docs/pretrained-models/ce-msmarco.html

Haystack's implementation uses this in its Ranker node. https://haystack.deepset.ai/pipeline_nodes/ranker

davidmezzetti:
The similarity pipeline is also using sequence classification. I'll take a look at what is going wrong. Most likely reason is the model outputs aren't interpreted correctly.

davidmezzetti:
Added a new pipeline type for cross-encoders. The similarity pipeline can use the cross-encoder pipeline as a reference if `crossencode=True`.

davidmezzetti:
It seems to be that the models work best with questions - longer passages with the answer (asymmetric similarity). For symmetric similarity, mnli models appear to work better. Definitely was worth adding.

PR #1195, opened 2026-08-22
fix: rank CrossEncoder by a fixed label, not top-1 position

### LINKED ISSUE #1189: Similarity/CrossEncoder(crossencode=True) ranks candidates by inconsistent labels with multi-class models (e.g. NLI cross-encoders)
### Describe the bug

`Similarity`/`CrossEncoder` with `crossencode=True` produces an inverted/meaningless ranking when the underlying cross-encoder model has more than one output label — e.g. any of the popular `cross-encoder/nli-*` models. The docs advertise general cross-encoder support (no caveat about label count), and NLI cross-encoders are a very natural thing to reach for since `Similarity` is framed as "ranks texts by relevance to a query" and NLI models are one of the most common cross-encoder types on the Hub.

### Root cause

In `src/python/txtai/pipeline/text/crossencoder.py`:

```python
result = self.pipeline([{"text": q, "text_pair": t} for t in texts], top_k=None, function_to_apply="none", num_workers=workers)
scores.append(self.function([r[0]["score"] for r in result], multilabel))
```

`top_k=None` makes the underlying HF `text-classification` pipeline return all labels **sorted by score, descending** — not in `id2label` order. Taking `r[0]["score"]` for every item therefore grabs "the score of whichever label this particular pair scored highest on", not a fixed label compared consistently across all candidates. This is correct by accident for single-output reranker models (the original use case in #372, e.g. `sentence-transformers/msmarco-MiniLM-*`, which only ever have one label) but silently breaks for any multi-class model.

### To Reproduce

```python
from txtai.pipeline import Similarity
from transformers import pipeline

premise = "There's nothing like the trendy resort clothing available here, styled on the island by the designers of the Ad-Lib group."
entailment_hyp = "The Ad-Lib group designs the trendy resort clothing here."
contradiction_hyp = "There is no trendy resort clothing available here, so you would have to go elsewhere to find it."

# Raw model — correct
clf = pipeline("text-classification", model="cross-encoder/nli-deberta-v3-base", top_k=None)
print(clf({"text": premise, "text_pair": entailment_hyp}))
# [{'label': 'entailment', 'score': 0.9984}, {'label': 'neutral', 'score': 0.0015}, {'label': 'contradiction', 'score': 0.00008}]
print(clf({"text": premise, "text_pair": contradiction_hyp}))
# [{'label': 'contradiction', 'score': 0.99995}, {'label': 'neutral', 'score': 0.00004}, {'label': 'entailment', 'score': 0.00001}]

# txtai Similarity — inverted
sim = Similarity(path="cross-encoder/nli-deberta-v3-base", crossencode=True)
print(sim(premise, [entailment_hyp, contradiction_hyp]))
# [(1, 0.9993...), (0, 0.9920...)]  <- ranks the CONTRADICTION above the ENTAILMENT
```

### Expected behavior

The contradiction hypothesis should not outrank the entailment hypothesis for a "relevance/similarity" ranking. At minimum the two candidates should be scored on the same, fixed label so the comparison is apples-to-apples.

### Suggested direction (open to a different approach)

Since there's no single universally-"correct" label to pick for an arbitrary N-class model, maybe an optional parameter (e.g. `label=` — index or name, defaulting to `None`/current behavior to stay backward compatible for single-label rerankers) so callers of multi-class cross-encoders can pick which class score to rank by? Happy to send a PR once the preferred direction is confirmed — didn't want to guess at API design and land something that doesn't fit.

### Environment

- txtai 9.12.0
- transformers (pipeline `text-classification`, `top_k=None`)
- Model used to reproduce: `cross-encoder/nli-deberta-v3-base`

St4r4x:
Fixes #1189.

## Problem
`CrossEncoder.__call__` (used by `Similarity(crossencode=True)`) always used `result[0]["score"]` per candidate - the top HF pipeline output, sorted by score descending. For a single-label reranker model that's the only score there is, so it's correct by construction (the original #372 use case). For a multi-class model (e.g. `cross-encoder/nli-*`), `result[0]` is whichever label that particular candidate happened to score highest on, so candidates end up compared on different, inconsistent labels - which is what inverts the ranking in the issue.

## Fix
Added an optional `labels=` parameter on `CrossEncoder`/`Similarity`, resolved by name/id via the model config - the same matching `Labels.limit()` already does. When given, every candidate is scored on that fixed label instead of its own top-1. `labels=None` (default) keeps the exact previous behavior.

`Labels.limit()` is now a `@staticmethod` taking `config` explicitly, so `CrossEncoder` calls the same label-matching logic instead of reimplementing it.

## Tests
- `testCrossEncoderLabels`: reproduces the inverted-ranking case with a multi-class NLI model (`prajjwal1/bert-medium-mnli`, already used elsewhere in this test file) and shows `labels=["1"]` (its entailment class) fixes it.
- `testCrossEncoderLabelsInvalid`: a `labels=` value matching no model label raises `ValueError` instead of silently scoring the wrong thing.
- Existing `testCrossEncoder`/`testCrossEncoderBatch` (single-label reranker, no `labels`) and all of `testlabels.py` pass unchanged, confirming the `labels=None` path is unaffected.
- Ran the full `testpipeline.testtext.testsimilarity` + `testpipeline.testtext.testlabels` modules and `pylint` (repo config) locally before opening this - no new warnings.

@sainikhiljuluri thanks for holding off, and for the precise write-up - implemented exactly as described above.

> [!NOTE]
> AI-assisted (Claude Code); verified against the source and a full local test + pylint run before posting.

### DIFF
--- a/src/python/txtai/pipeline/text/crossencoder.py
+++ b/src/python/txtai/pipeline/text/crossencoder.py
@@ -7,6 +7,8 @@ from ..hfpipeline import HFPipeline
 # Core library imports
 from ...util import Library
 
+from .labels import Labels
+
 np = Library().numpy()
 
 
@@ -18,7 +20,7 @@ class CrossEncoder(HFPipeline):
     def __init__(self, path=None, quantize=False, gpu=True, model=None, **kwargs):
         super().__init__("text-classification", path, quantize, gpu, model, **kwargs)
 
-    def __call__(self, query, texts, multilabel=True, workers=0):
+    def __call__(self, query, texts, multilabel=True, workers=0, labels=None):
         """
         Computes the similarity between query and list of text. Returns a list of
         (id, score) sorted by highest score, where id is the index in texts.
@@ -32,6 +34,10 @@ class CrossEncoder(HFPipeline):
             texts: list of text
             multilabel: labels are independent if True, scores are normalized to sum to 1 per text item if False, raw scores returned if None
             workers: number of concurrent workers to use for processing data, defaults to None
+            labels: optional list of labels to score by, resolved by name/id via the model config the same way
+                    Labels.limit() does. Restricts ranking to a fixed label so multi-class models (e.g. NLI
+                    cross-encoders) compare every candidate on the same class, instead of whichever one each
+                    candidate happened to score highest on. Defaults to that original top-scored label when None.
 
         Returns:
             list of (id, score)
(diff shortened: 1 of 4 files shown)


davidmezzetti:
Thanks for the PR.

What if instead of making `limit` method static and sharing it, we just had `CrossEncoder` extend `Labels` like how `Similarity` does? You can add `**kwargs` in a similar fashion.

St4r4x:
Good call - done. `CrossEncoder` now extends `Labels` and reuses `limit()` directly as an inherited method, same relationship `Similarity` already has - smaller diff, `labels.py` is untouched.

Re-verified after the change: full local test suite + pylint still green, and re-ran the real-model repro from the issue (`cross-encoder/nli-deberta-v3-base`) - identical output to before the refactor.

davidmezzetti:
LGTM - Thank you!

PR #1196, opened 2026-08-22
Resolve LEMUR and MUVERA settings

morgan-coded:
`lemur` and `muvera` values that aren't dictionaries reach the constructor splats in `LatePooling` and raise a raw `TypeError`, so `muvera: false` fails instead of disabling MUVERA. This resolves `None` and `false` as disabled for either encoder, a bare `lemur` string as the artifact path, and `muvera: true` as the constructor defaults, while anything else raises a `ValueError` naming the key and the accepted forms. Documented dictionary forms construct exactly what they did, and `lemur` still takes precedence when both are set. I added tests for each form and ran the full `testpooling` module with 22 passing, plus pylint and black.


davidmezzetti:
LGTM - Thank you!

PR #1198, opened 2026-08-23
Fix WordVectors.index() silently dropping the checkpoint argument

### LINKED ISSUE #1197: WordVectors.index() accepts a checkpoint parameter but never uses it
`WordVectors.index()` (the `words` backend, used for fastText/word2vec style vectors) accepts a `checkpoint` parameter but never uses it, so indexing resume is silently a no-op for this backend even though it is enabled and works for every other backend.

## Reproduction

```python
import os, tempfile
from txtai.vectors import VectorsFactory

checkpoint = tempfile.mkdtemp()
documents = [(x, "This is a test", None) for x in range(10)]

model = VectorsFactory.create({"path": "neuml/glove-6B-quantized", "parallel": False}, None)
model.index(documents, 5, checkpoint)

print(os.path.exists(os.path.join(checkpoint, model.vectorsid())))
```

Expected: `True` (a checkpoint file should exist after indexing with `checkpoint` set, same as `Vectors.index()`). Actual on current `main`: `False`. Same result with `parallel: True`.

Confirmed on current `main` in a clean container, both the serial and parallel code paths, plus a kill-and-resume scenario: interrupting an indexing run partway through and resuming with the same `checkpoint` directory re-embeds every document from scratch instead of skipping the batches already written before the interruption.

## Root cause

`WordVectors.index()` (`src/python/txtai/vectors/dense/words.py`) has two branches:

- The non-parallel branch calls `super().index(documents, batchsize)`, dropping the `checkpoint` argument entirely, so `Vectors.index()`'s own `Recovery` wiring never sees it.
- The parallel (multiprocessing `Pool`) branch writes straight to a plain `tempfile.NamedTemporaryFile` and never touches `Recovery` at all.

`Vectors.index()` in `vectors/base.py` wires `Recovery` correctly and every other vector backend routes through it unchanged; `WordVectors` is the only backend that overrides `index()`, and it is the only one where this is broken.

AmirF194:
## Root cause

`WordVectors.index()` (`src/python/txtai/vectors/dense/words.py`) accepts a `checkpoint` argument but never uses it. The non-parallel branch calls `super().index(documents, batchsize)`, dropping `checkpoint` before it reaches `Vectors.index()`'s own `Recovery` wiring. The parallel (multiprocessing `Pool`) branch writes straight to a plain `tempfile.NamedTemporaryFile` and never touches `Recovery` at all. `WordVectors` is the only backend that overrides `index()`, so every other backend already gets checkpointing for free through `Vectors.index()`; this one silently drops it in both of its own code paths.

## Fix

- Non-parallel branch: pass `checkpoint` through to `super().index()`.
- Parallel branch: mirror `Vectors.index()`'s pattern. Generate a `Recovery` instance when `checkpoint` is set, write through `self.spool(checkpoint, vectorsid)` instead of a bare temp file, and batch documents so each batch can be served from `recovery()` before falling back to the pool. Extracted the per-batch logic into `parallelbatch()` since the outer loop needed to call it twice (mid-loop flush and final partial batch), same shape as the existing non-parallel `Vectors.batch()`.

## Tests

Added `testIndexCheckpoint` (serial) and `testIndexCheckpointParallel` (parallel) to `test/python/testvectors/testdense/testwordvectors.py`. Both fail on `main` and pass on this branch:

- `testIndexCheckpoint` asserts the checkpoint file exists after indexing, then re-indexes the same documents and asserts `encode` is never called again (fully served from the checkpoint).
- `testIndexCheckpointParallel` asserts the checkpoint file exists for the parallel path (the second assertion cannot cross the `multiprocessing.Pool` fork boundary reliably, so I checked it separately, see below).

## Verified

- New tests fail on `main` (`a10667a1`) and pass on this branch, in Docker.
- Kill-and-resume checked directly for both branches: interrupted an indexing run after 10 of 20 documents, then resumed with the same checkpoint directory over the full document list. On `main`, the second run re-embeds all 20. On this branch, it re-embeds only the 10 that were not yet checkpointed, for both `parallel: True` and `parallel: False`.
- `test/python/testvectors/testdense/testwordvectors.py` and `testvectors.py`: all pass, using the real `neuml/glove-6B-quantized` model.
- `black --check` and `pylint` (10.00/10, repo's own `.pylintrc`) plus `pre-commit run` (both hooks) on the changed files.
- Coverage on the changed package: every line touched by this diff is exercised by the test file above; did not run the full `make data coverage` matrix (needs the full `[all,dev]` extras and downloaded fixtures), scoped to the changed module and its tests.

Fixes #1197

### DIFF
--- a/src/python/txtai/vectors/dense/words.py
+++ b/src/python/txtai/vectors/dense/words.py
@@ -5,7 +5,6 @@ Word Vectors module
 import json
 import logging
 import os
-import tempfile
 
 from multiprocessing import Pool
 
@@ -21,6 +20,7 @@ from ...pipeline import Tokenizer
 from ...util import Download, DownloadError, Library
 
 from ..base import Vectors
+from ..recovery import Recovery
 
 # Core library imports
 np = Library().numpy()
@@ -158,7 +158,7 @@ class WordVectors(Vectors):
 
         # Use default single process indexing logic
         if not parallel:
-            return super().index(documents, batchsize)
+            return super().index(documents, batchsize, checkpoint)
 
         # Customize indexing logic with multiprocessing pool to efficiently build vectors
         ids, dimensions, batches, stream = [], None, 0, None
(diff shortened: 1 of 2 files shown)


davidmezzetti:
Would you mind syncing your branch with the latest? I think that will fix the build error.

AmirF194:
Synced with master and pushed. The build failure was an unrelated HF Hub lookup in a different test file, not the checkpoint fix itself, so this should clear it.

davidmezzetti:
LGTM - Thank you!

PR #1205, opened 2026-08-25
Fix safetensors numpy/torch submodule imports

### LINKED ISSUE #1204: [bug] ANN backends crash saving safetensors storage since 9.9.0 (regression)
## Summary

Saving an ANN index with safetensors storage crashes with `AttributeError: module 'safetensors' has no attribute 'numpy'` on a **default install** (`pip install txtai`, no extras). This is a regression introduced in **txtai 9.9.0** — 9.8.0 and earlier imported `safetensors.numpy` explicitly.

## Environment

* txtai: 9.9.0 – 9.13.0 (verified on 9.13.0, latest)
* Python: 3.10, 3.14 (version-independent)
* OS: Windows 11 (also reproduced on Linux in a clean container)
* Install mode: default `pip install txtai` (no extras)
* safetensors: 0.8.0 (also broken with 0.7.0)
* torch: 2.13.0

## Steps to reproduce

```bash
pip install txtai
```

```python
import os
import tempfile

import numpy as np

from txtai.ann import ANNFactory

ann = ANNFactory.create({"backend": "numpy", "numpy": {"safetensors": True}})
ann.index(np.random.rand(100, 240).astype(np.float32))
ann.save(os.path.join(tempfile.gettempdir(), "index.safetensors"))
```

The official test `test/python/testann/testdense.py::testNumPySafetensors` fails with the same error (reproduced 5/5).

## Expected behavior

The index is saved to `index.safetensors`.

## Actual behavior

```text
Traceback (most recent call last):
  ...
  File ".../txtai/ann/dense/numpy.py", line 95, in save
    self.savesafetensors({"data": self.numpy(self.backend)}, path)
  File ".../txtai/ann/dense/numpy.py", line 175, in savesafetensors
    safetensors.numpy.save_file(data, path, metadata)
AttributeError: module 'safetensors' has no attribute 'numpy'
```

## Root cause

* `numpy.py` line 14 loads safetensors through the lazy helper `library.safetensors()`, which only executes `import safetensors`.
* Line 175 then calls `safetensors.numpy.save_file(...)` — the `numpy` submodule attribute only exists if **some other package in the process imported `safetensors.numpy` first**.
* The `safetensors` package itself never auto-imports its `numpy`/`torch` submodules (checked `__init__.py` in 0.7.0 and 0.8.0).
* Regression history:
  * 9.8.0 and earlier: `from safetensors.numpy import save_file` (explicit import) — works
  * 9.9.0: refactored to `library.safetensors()` + `safetensors.numpy.save_file` — broken
* **Why CI is green:** the build installs `.[all,dev]`, and the `staticvectors` package (in the `vectors` extra) happens to `import safetensors.numpy` at package import time, which masks the missing import in every CI run.

## Additional affected paths

* `src/python/txtai/ann/dense/torch.py` — `savesafetensors()` delegates to `super().savesafetensors()` (`numpy.py`), so the Torch backend (the default for `Embeddings`) hits the same error whenever safetensors storage is used (e.g. quantization config auto-sets `safetensors: true`, see `torch.py` line 59-60).
* `src/python/txtai/models/pooling/lemur.py` line 196 — same pattern with `safetensors.torch.save_file(...)` (the `torch` submodule is only present because `transformers` imports it; nothing in the default install imports `safetensors.numpy`).

## Suggested fix

Restore the explicit submodule imports (2-3 lines):

```python
# txtai/ann/dense/numpy.py
import safetensors.numpy  # or: from safetensors.numpy import save_file

# txtai/models/pooling/lemur.py
import safetensors.torch
```

The existing test `testNumPySafetensors` covers the numpy path; a small addition would cover the torch/lemur save path.

Evhye38496:
---

Fixes #1204

## Root cause

`txtai/ann/dense/numpy.py` loads safetensors through the lazy helper `library.safetensors()`, which only executes `import safetensors`. Line 175 then calls `safetensors.numpy.save_file(...)`, but the `numpy` submodule attribute only exists if **some other package in the process imported `safetensors.numpy` first** — safetensors never auto-imports its `numpy`/`torch` submodules (checked `__init__.py` in 0.7.0 and 0.8.0).

This is a regression introduced in **9.9.0**:

* 9.8.0 and earlier: `from safetensors.numpy import save_file` (explicit import) — works
* 9.9.0: refactored to `library.safetensors()` + `safetensors.numpy.save_file` — broken

It is masked in CI because the build installs `.[all,dev]`, and `staticvectors` (in the `vectors` extra) happens to `import safetensors.numpy` at package import time.

## Change

`library.safetensors()` now explicitly imports the `safetensors.numpy` and `safetensors.torch` submodules. A single location covers every affected path:

* NumPy ANN backend (`numpy.py`)
* Torch ANN backend (inherits `savesafetensors` via `super()`)
* LEMUR pooling model save (`models/pooling/lemur.py`)

The existing `ImportError` fallback stub is unchanged, so installs without safetensors degrade the same way as before.

## Tests

* `test/python/testann/testdense.py::testNumPySafetensors` — fails on `master` with `AttributeError: module 'safetensors' has no attribute 'numpy'`, passes with this change
* Full `testdense.py` suite: no new failures (remaining errors are pre-existing optional-dependency skips)
* `black --line-length 150 --check` and `pylint --rcfile=.pylintrc` (10.00/10) clean on the changed file
* Verified the stub fallback path (safetensors not installed) still raises the expected `ImportError` on attribute access

### DIFF
--- a/src/python/txtai/util/library.py
+++ b/src/python/txtai/util/library.py
@@ -212,6 +212,10 @@ class Library:
         try:
             import safetensors
 
+            # Import submodules - they are not auto-imported with safetensors
+            import safetensors.numpy
+            import safetensors.torch
+
         except ImportError:
 
             class Safetensors:


davidmezzetti:
LGTM - Excellent catch!

PR #1210, opened 2026-08-28
Fix bash tool hanging on commands that read stdin

### LINKED ISSUE #1209: Bash tool hangs forever on an allowed command that reads stdin
The bash tool runs commands with `subprocess.run(command, capture_output=True, text=True, check=False)` in `src/python/txtai/agent/tool/bash.py`. `capture_output=True` pipes stdout and stderr but leaves stdin inherited from the parent.

Every command in the default allowed list (`cat`, `cut`, `diff`, `grep`, `head`, `tail`) reads stdin when called with no file operand, so `agent.tools["bash"](["cat"])` blocks on a read that never returns. In a long-lived parent such as an API service or a notebook, stdin never reaches EOF, and since no `timeout` is passed there is nothing to break the wait — the agent is wedged permanently.

Reproduction against master:

```python
from txtai.agent.tool import ToolFactory
ToolFactory.default("bash")(["cat"])   # never returns
```

Omitting a file argument is an ordinary model slip, not an adversarial input.

Suggested fix: pass `stdin=subprocess.DEVNULL` so such a command returns immediately, and add a configurable `timeout` so any other slow command is bounded. Happy to open a PR.

pacocartones:
## Problem

`BashTool.forward` ran the command with no timeout and no stdin redirection:

```python
output = subprocess.run(command, capture_output=True, text=True, check=False).stdout
```

`capture_output=True` sets `stdout` and `stderr` to pipes but leaves `stdin` alone, so the child inherits the parent's stdin.

Every command in the default allowed list — `cat`, `cut`, `diff`, `grep`, `head`, `tail` — reads stdin when called with no file operand. So a model emitting `["cat"]`, or `["grep", "pattern"]` with the path omitted, blocks on a read that never returns. In a long-lived parent such as an API service or a notebook, stdin never reaches EOF, and because no `timeout` is passed there is nothing to break the wait: the agent is wedged permanently, holding the thread and the step budget.

Reproduced against `master` with the tool's real `forward()`:

```
REPRODUCED: BashTool.forward(['cat']) STILL BLOCKED after 10s
```

A model does not have to be adversarial to reach this — omitting a file argument is an ordinary generation slip, and the tool description invites `cat`.

## Fix

- Redirect stdin to `DEVNULL`, so a command that reads stdin sees EOF immediately and returns empty output instead of blocking.
- Add a `timeout` constructor argument, defaulting to 30 seconds, so a command that is slow for any other reason reports a timeout rather than hanging.

Both are needed. `DEVNULL` alone leaves any other long-running command unbounded, and a timeout alone would make every stdin-reading command wait the full timeout before failing.

The timeout returns a message rather than raising, so the agent can read what happened and adapt on its next step.

`ToolFactory.default("bash")` keeps working unchanged — both arguments are optional.

## Tests

`testToolsBashStdin` replaces stdin with a pipe that never reaches EOF, simulating a long-lived parent, then calls `["cat"]` on a thread and asserts it returns. On `master` it fails after blocking the full 30s join:

```
FAIL: testToolsBashStdin
AssertionError: True is not false
```

`testToolsBashTimeout` asserts a command exceeding the timeout reports one. On `master` it errors, as there is no `timeout` argument:

```
TypeError: BashTool.__init__() got an unexpected keyword argument 'timeout'
```

Both pass with the change.

- `python -m unittest -v testagent` — 11 tests, OK
- `black --check` clean, `pylint` 10.00/10

Closes #1209.

### DIFF
--- a/src/python/txtai/agent/tool/bash.py
+++ b/src/python/txtai/agent/tool/bash.py
@@ -6,6 +6,9 @@ import subprocess
 
 from smolagents import Tool
 
+# Default maximum time in seconds to wait for a command to complete
+TIMEOUT = 30
+
 
 class BashTool(Tool):
     """
@@ -14,12 +17,13 @@ class BashTool(Tool):
     """
 
     # pylint: disable=W0231
-    def __init__(self, allowed=None):
+    def __init__(self, allowed=None, timeout=None):
         """
         Creates a BashTool.
 
         Args:
             allowed: list of allowed commands to run, has limited set of defaults which are a best effort not a sandbox
+            timeout: maximum number of seconds to wait for a command to complete, defaults to 30
         """
 
         # Tool parameters
@@ -33,6 +37,9 @@ class BashTool(Tool):
         # Default list of allowed commands
         self.allowed = allowed if allowed else ["cat", "cut", "diff", "grep", "head", "ls", "tail"]
 
+        # Maximum time to wait for a command
+        self.timeout = timeout if timeout else TIMEOUT
+
         # Validate parameters and initialize tool
         super().__init__()
 
@@ -40,6 +47,9 @@ class BashTool(Tool):
     def forward(self, command):
         """
         Runs a shell command as a subprocess.
+
+        stdin is closed and a timeout is applied. Otherwise an allowed command called with no file arguments
+        (i.e. `cat`) reads from stdin and blocks the agent indefinitely.
 
         Args:
             command: command arguments as a list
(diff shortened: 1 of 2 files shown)


davidmezzetti:
Nice catch, thank you for the PR. Can we make a couple minor changes.

```diff
-# Default maximum time in seconds to wait for a command to complete
-TIMEOUT = 30
-  

-def __init__(self, allowed=None, timeout=None):
+def __init__(self, allowed=None, timeout=30):

-self.timeout = timeout if timeout else TIMEOUT
+self.timeout = timeout
```

I'm fine with the 30s default but this way we'd still let someone turn off the timeout if they prefer.

davidmezzetti:
@pacocartones Still interested in completing this PR?

pacocartones:
Thanks for the review. Updated as suggested: the module-level `TIMEOUT` constant is gone, `timeout=30` is now the default argument of `BashTool.__init__` and is assigned directly, so passing `timeout=None` disables the timeout entirely (`subprocess.run` treats `None` as no limit). The existing timeout test also asserts that `BashTool(timeout=None)` leaves `timeout` as `None`; black and pylint are clean.

davidmezzetti:
LGTM - Thank you!

PR #1212, opened 2026-09-01
Add ONNX vectors backend

### LINKED ISSUE #1211: Add ONNX vectors backend
Currently, `txtai` supports ONNX models via a custom model that emulates a Torch model. While this works, it requires a torch install which defeats the purpose of supporting ONNX. It would be best to add a new vectors backend that supports ONNX directly like the other available backends (i.e. LiteRT, llama.cpp etc). 

The Torch ONNX interface should otherwise stay untouched as it's also used by the pipeline framework which still requires Torch.

joseph-bergin:
Closes #1211

Adds a dense vectors backend that runs ONNX models directly through `onnxruntime`, so ONNX embeddings no longer require the Torch install that the existing `models/onnx.py` wrapper depends on. The Torch ONNX interface under `pipeline` is untouched.

### Implementation

`vectors/dense/onnx.py` follows the `litert.py` layout: conditional import, `ismodel()` matching `.onnx`, `loadmodel()` and `encode()`. Registered in `VectorsFactory` as method `onnx`, with path inference alongside the other backends, and `onnxruntime>=1.11.0` added to the vectors extra to match the pin used elsewhere in setup.py.

Following your notes on the issue:

- **Dynamic shapes.** No static batch buffers as in LiteRT. Batching is driven by `encodebatch` and the tokenizer pads to the longest sequence per batch.
- **Reuse from `models/onnx.py`.** `inputs()` sets only the inputs the session declares, the same check `OnnxModel.parse` makes, since not every export takes `token_type_ids`. `providers()` prefers CUDA when available, but tests `onnxruntime.get_available_providers()` on its own rather than going through `torch.cuda.is_available()`, since requiring Torch here would defeat the purpose.

Two things worth a look:

- **Pooling.** Token-level output is mean pooled against the attention mask so padding is excluded; output that is already 2-D passes through untouched. That covers both `last_hidden_state` exports and pooled exports, but let me know if you would rather this only accept pooled models and leave pooling to the caller.
- **Tokenizer resolution.** Reads the `tokenizer` config option first, matching how ONNX models are already configured, and otherwise falls back to a `tokenizer.json` next to the model the way LiteRT does.

### Testing

`test/python/testvectors/testdense/testonnx.py` exports `paraphrase-MiniLM-L3-v2` with `HFOnnx`, mirroring `testpipeline/testtrain/testonnx.py::testPooling`, then covers factory method resolution, indexing and dimensions, stability when a batch spans multiple encode batches, and that pooled embeddings keep their semantics.

To be straight about what I could run: my environment has no route to the HF Hub, so I could not execute that test locally. What I did verify was the logic itself, against a synthetic ONNX graph and tokenizer built offline:

- 3-D token output mean pooled to one vector per input
- models that declare `token_type_ids` and models that do not, both fed correctly
- already-pooled 2-D output passing through
- a padded short input producing the same vector as that input alone, so padding does not leak into the mean
- identical results with `encodebatch` at 2 and at 32

`black --line-length 150` and `pylint --rcfile=.pylintrc` are clean (10.00/10) on all changed files.

AI assistance was used for this change.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -142,6 +142,7 @@ extras["vectors"] = [
     "litellm>=1.37.16",
     "llama-cpp-python>=0.2.75",
     "model2vec>=0.3.0",
+    "onnxruntime>=1.11.0",
     "scikit-learn>=0.23.1",
     "scipy>=1.4.1",
     "sentence-transformers>=5.0.0",
--- a/src/python/txtai/vectors/dense/factory.py
+++ b/src/python/txtai/vectors/dense/factory.py
@@ -11,6 +11,7 @@ from .litellm import LiteLLM
 from .litert import LiteRT
 from .llama import LlamaCpp
 from .m2v import Model2Vec
+from .onnx import ONNX
 from .sbert import STVectors
 from .words import WordVectors
 
@@ -57,6 +58,10 @@ class VectorsFactory:
         if method == "model2vec":
             return Model2Vec(config, scoring, models)
 
+        # ONNX vectors
+        if method == "onnx":
+            return ONNX(config, scoring, models)
+
         # Sentence Transformers vectors
         if method == "sentence-transformers":
             return STVectors(config, scoring, models) if config and config.get("path") else None
@@ -99,6 +104,8 @@ class VectorsFactory:
                     method = "llama.cpp"
                 elif Model2Vec.ismodel(path):
                     method = "model2vec"
+                elif ONNX.ismodel(path):
+                    method = "onnx"
                 elif WordVectors.ismodel(path):
                     method = "words"
                 else:
--- a/src/python/txtai/vectors/dense/onnx.py
+++ b/src/python/txtai/vectors/dense/onnx.py
@@ -1,0 +1,160 @@
+"""
+ONNX module
+"""
+
+import os
+
+# Conditional import
+try:
+    import onnxruntime as ort
+
+    from tokenizers import Tokenizer
+
+    ONNX_RUNTIME = True
+except ImportError:
+    ONNX_RUNTIME = False
+
+from ...util import Download, Library
+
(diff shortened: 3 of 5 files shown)


davidmezzetti:
Hello - Can you make sure the tests cover the new untested code: https://coveralls.io/jobs/187192998

joseph-bergin:
Thanks, that was a fair gap. The exported pooling model returns a 2D output, is configured with a tokenizer and runs on CPU, so mean pooling, input filtering, tokenizer fallback, truncation and provider selection were never exercised.

Added `TestONNXModels`, which builds small ONNX graphs locally rather than downloading anything, covering:

- token level output pooled to one vector per input, and pooled output passing through untouched
- `token_type_ids` set only when the model declares it
- padding excluded from the mean, a padded input produces the same vector as that input alone
- `maxlength` truncation
- a `tokenizer.json` found next to the model when none is configured
- provider selection with and without CUDA available
- the error raised when `onnxruntime` isn't installed

That takes `onnx.py` to 95% locally, the remaining lines being the conditional import fallback that the other backends have too.

I also corrected the batching test from the first CI run. It asserted embeddings were bitwise close regardless of batch size, which isn't true since each input is padded to the longest sequence in its own batch. It compares cosine similarity now, and the export no longer quantizes so the comparison isn't dominated by int8 noise.

davidmezzetti:
Thanks. You can add tests for the optional fallback here: https://github.com/neuml/txtai/blob/master/test/python/testoptional.py#L361. It already removes onnxruntime.

That is how you can test out the missing deps.

joseph-bergin:
Done. Moved it to `testoptional.py::testVectors` alongside the other backends and dropped the patched version.

davidmezzetti:
LGTM - Thank you!

PR #1213, opened 2026-09-02
fix(api): parse the graph query parameter as a bool

serhiizghama:
`GET /search` takes `graph` straight off the query string and passes the raw value through, so any non-empty value turns graph mode on. `graph=false`, `graph=False` and `graph=0` all come back with a serialized graph instead of the usual result list. `POST /batchsearch` declares it as `graph: bool = Body(default=False)` and coerces it properly, so the two endpoints disagree on the same documented parameter.

Everything else in `API.search` is already parsed on the way in — `limit` goes through `self.limit()`, `weights` through `self.weights()`, `parameters` through `json.loads` — `graph` was the only one left raw. Added a `graph()` helper next to the others that accepts the same truthy values FastAPI does.

This also hits clusters quietly: `Cluster.search` appends `&graph={graph}` unconditionally, so with the default of `False` every shard request carried `graph=False` and a graph-enabled index ran in graph mode on every search.

The test starts a keyword + graph index through the API and checks that `graph=true` returns a graph while `false`, `False` and `0` return a list. It fails on main for those three.

### DIFF
--- a/src/python/txtai/api/base.py
+++ b/src/python/txtai/api/base.py
@@ -32,7 +32,7 @@ class API(Application):
         weights = self.weights(request.query_params.get("weights") if request and hasattr(request, "query_params") else weights)
         index = request.query_params.get("index") if request and hasattr(request, "query_params") else index
         parameters = request.query_params.get("parameters") if request and hasattr(request, "query_params") else parameters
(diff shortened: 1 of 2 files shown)


davidmezzetti:
LGTM - Thank you!

PR #1214, opened 2026-09-02
Load the Milvus collection when opening a saved index

morgan-coded:
With `milvus-lite` 3.2.1, `search()` failed after `save()` or `load()` because an indexed collection reopened in the `released` state, while `count()` still worked. `open()` now calls `load()` on the reopened collection, and the idempotent collection load leaves fresh index behavior unchanged. I reproduced the released-collection failure, extended the shared save helper to search after save and reload, then ran `testann.testdense`, which completed 35 tests with 3 skipped and no failures, plus pre-commit.


davidmezzetti:
LGTM - Thank you!

PR #1215, opened 2026-09-02
Ignore padding rows in LateEncoder scores

morgan-coded:
`LateEncoder` scores changed with the padded query length, and zero-padded document rows could win the per-token maximum under centering. This masks padded query and document rows before averaging, so single-query scores and rankings without a padding floor stay unchanged; it applies the same padding mechanism fixed for MUVERA in #1177. I reproduced both padding paths, added focused batch and exact-MaxSim checks, then ran `testpipeline.testtext.testsimilarity` with 11 tests passing and pre-commit.


davidmezzetti:
LGTM - Thank you!

PR #1216, opened 2026-09-02
Resolve label filters by label id

morgan-coded:
`Labels.limit` resolved requested labels by their position in `label2id` instead of their id, so models with a different insertion order returned the wrong class. This maps names and numeric strings directly to label ids, so `CrossEncoder.score` inherits the correction while `labels=None` and unmatched labels stay unchanged; it follows the label-filter path added in #1195. I added fixed-label and cross-encoder ordering tests, then ran `testpipeline.testtext.testlabels` with 7 tests passing, `testpipeline.testtext.testsimilarity` with 11 passing, and pre-commit.

### DIFF
--- a/src/python/txtai/pipeline/text/labels.py
+++ b/src/python/txtai/pipeline/text/labels.py
@@ -120,19 +120,17 @@ class Labels(HFPipeline):
         result = [(config.label2id.get(x["label"], 0), x["score"]) for x in result]
 
         if labels:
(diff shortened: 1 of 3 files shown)


davidmezzetti:
Interesting! Certainly the right way to handle it but the labels must be in order more often than not.

davidmezzetti:
LGTM - Thank you!

PR #1218, opened 2026-09-02
Keep count() accurate after repeated deletes in GGML and IVFSparse

morgan-coded:
GGML and `IVFSparse` `count()` drifted low after repeated deletes, and `IVFSparse` also counted unknown ids as deletions; the bad counts survived save and load. Both backends now ignore ids already marked deleted, `IVFSparse` ignores ids that were never indexed, and search behavior stays unchanged; this is the same delete-accounting shape as #1202. I added repeat-delete and reload coverage, then ran `testann.testdense` with 36 tests, 3 skipped, `testann.testsparse` with 7 tests passing, and pre-commit.

### DIFF
--- a/src/python/txtai/ann/dense/ggml.py
+++ b/src/python/txtai/ann/dense/ggml.py
@@ -220,9 +220,10 @@ class GGMLTensors:
 
         shape = utils.get_shape(self.data)
 
-        # Filter any index greater than size of array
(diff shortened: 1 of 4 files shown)


davidmezzetti:
Nice catch! Looks like a similar issue to this one: #1202

davidmezzetti:
LGTM - Thank you!

PR #1219, opened 2026-09-03
Delete every index entry for an id in Terms

morgan-coded:
`Terms.delete` marked only the first index entry for an id, so delete, upsert, and delete again left the live copy behind and the next matching `search()` raised `KeyError`. It now marks every index entry for the requested ids, while unknown ids and repeated deletes remain idempotent; this extends the repeated-delete handling from #1202. I added the cycle across BM25, TF-IDF, and SIF, then ran `testscoring.testkeyword` with 12 tests passing and pre-commit.

### DIFF
--- a/src/python/txtai/scoring/terms.py
+++ b/src/python/txtai/scoring/terms.py
@@ -129,14 +129,13 @@ class Terms:
             ids: ids to delete
         """
 
-        # Set index ids as deleted, ignoring ids that were never indexed and ids
-        # already marked deleted. Skipping duplicates keeps count() accurate when
-        # the same id is deleted more than once (e.g. repeated upsert/delete flows).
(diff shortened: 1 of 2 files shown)


davidmezzetti:
Good catch on this. Although wouldn't iterating over the entire `self.ids` get quite slow for large indexes? 

Perhaps something that iterates through using `index()`? Then uses the start index until it can't find anymore? That would enable C loop processing which is much faster.

morgan-coded:
Switched `Terms.delete` to walk `self.ids.index()` forward from each match for every requested id. The existing reinsert, idempotent, and unknown-id cases still cover the behavior.

davidmezzetti:
LGTM - Thank you!

PR #1220, opened 2026-09-03
fix(vectors): remove the sparse spool file when checkpointing is disabled

serhiizghama:
`SparseVectors.vectors()` overrides the base implementation but doesn't carry over its cleanup step, so the `.npy` spool file that `index()` writes is left behind on every call. `Vectors.vectors()` ends with `if not checkpoint: os.remove(stream)`; the sparse version just returns. Since `spool()` uses `NamedTemporaryFile(delete=False)`, nothing else removes it — every sparse index build leaves a full copy of the serialized embeddings in the temp dir, and `Sparse.encode()` in the scoring path goes through here too.

Added the same guarded remove, so the checkpoint case still keeps its file for restart. Verified both paths locally: with a checkpoint directory the file is preserved and a second run is served from it, without one the temp dir is empty afterwards. New test spools into an empty directory and asserts nothing is left; it fails on master with `['tmp7dloe8t7.npy'] != []`.

### DIFF
--- a/src/python/txtai/vectors/sparse/base.py
+++ b/src/python/txtai/vectors/sparse/base.py
@@ -2,6 +2,8 @@
 SparseVectors module
 """
 
+import os
+
 # Conditional import
 try:
     from scipy.sparse import csr_matrix, vstack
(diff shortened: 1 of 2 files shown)


davidmezzetti:
Good catch!

davidmezzetti:
LGTM - Thank you!

PR #1222, opened 2026-09-03
Keep the archive working directory alive when saving a loaded index

morgan-coded:
An index loaded from a `.tar.gz` or `.zip` archive could not be saved again to the same archive, another archive, or a directory when keyword or content storage kept SQLite connections open in the extracted working directory. This keeps the current archive working directory alive across `save()` and `exists()`, while `load()` still resets it when replacing the index; fresh non-archive indexes and dense-only archive behavior stay unchanged. I added a focused test covering save-back, `exists()`, save-to-directory, `upsert()`, and reload, then ran it with pre-commit and the existing embeddings tests; the new test passes and nothing else changed.

### DIFF
--- a/src/python/txtai/embeddings/base.py
+++ b/src/python/txtai/embeddings/base.py
@@ -549,6 +549,9 @@ class Embeddings:
         if cloud:
             path = cloud.load(path)
 
+        # Reset archive since this replaces the current index
+        self.archive = None
+
         # Check if this is an archive file and extract
         path, apath = self.checkarchive(path)
         if apath:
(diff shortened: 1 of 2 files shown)


davidmezzetti:
LGTM - Thank you!

In case you're curious on when I pick "Create a Merge commit" and "Squash and Merge" it has to do with the verbosity of the commit messages. A lot of coding agents write a novel for each commit and I don't want that mucking up the commit log. So in those cases I squash and merge into a single commit. For simple commit logs I merge that directly in. Doesn't matter either way but just an FYI.

If I want more info I can look at the git diff and the GitHub Issue/PR which normally has a ton of information, especially if it's from coding agent. It writes a novel justifying why I have to integrate the PR.

PR #1223, opened 2026-09-03
Key the vectors model cache by configuration

morgan-coded:
Subindexes using the same encoder path shared one cached model, so later subindexes silently ignored their own `vectors` settings. This keys the cache by the path and load-time configuration, while identical configurations still share a single model load. I reproduced the cross-subindex reuse, added a focused MUVERA settings test, then ran the Hugging Face vectors module with 5 tests passing and pre-commit.

### DIFF
--- a/src/python/txtai/vectors/base.py
+++ b/src/python/txtai/vectors/base.py
@@ -96,9 +96,12 @@ class Vectors:
             model
         """
 
-        # Cache key is the model path plus settings that change the loaded model. Other settings share a single load.
-        config = {"path": path, "method": self.config.get("method"), "vectors": self.config.get("vectors", {})}
-        key = json.dumps(config, sort_keys=True, default=str)
+        # Build a cache key from settings that change the loaded model when a configuration is present.
+        if self.config:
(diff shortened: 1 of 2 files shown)


davidmezzetti:
The original intent behind this caching was to prevent loading the same heavy torch model even when there is different configuration for the underlying model. I think that was probably heavy handed in some cases but there is now the case where the same model will be loaded in memory twice just for a different maxlength.

morgan-coded:
Narrowed the cache key to the path, method and `vectors` settings, so changes to `maxlength`, `tokenizer`, `gpu` or `instructions` share the loaded model as before. Different `vectors` settings still load separately because they change the vectors produced. The test now covers both sharing and separation. I can follow up by caching the underlying model once and building the pooling wrapper per subindex if you would rather take that approach.

davidmezzetti:
Looks like this change needs to wrap a check for `if self.config`. That's what the failing test is from.

morgan-coded:
Guarded the cache-key build with `if self.config`, so a `Vectors` instance created without a configuration reaches `loadmodel` as before. The `testvectors` module passes again.

davidmezzetti:
LGTM - Thank you!

PR #1224, opened 2026-09-03
Save empty keyword and content indexes without errors

morgan-coded:
`index([])` followed by `save()` raised `AttributeError` for keyword or hybrid scoring and for `content: True` because the SQLite connection is created on first insert. This initializes the schemas during `save()` and writes the empty index-id mapping, so an empty keyword index with or without content reloads with `count()` returning 0, while populated indexes save exactly as before; it completes the empty Terms save side of #1143's search fix. I added focused empty-save tests, then ran the keyword module with 12 tests passing and pre-commit.

### DIFF
--- a/src/python/txtai/database/embedded.py
+++ b/src/python/txtai/database/embedded.py
@@ -31,6 +31,9 @@ class Embedded(RDBMS):
         self.path = path
 
     def save(self, path):
+        # Initialize connection if not open
+        self.initialize()
+
         # Temporary database
         if not self.path:
             # Save temporary database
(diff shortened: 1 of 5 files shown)


davidmezzetti:
What would be the use case for saving an empty index? Is this just for consistency with the ANN which doesn't crash in this case?

morgan-coded:
Mostly consistency, yes: the concrete case is an indexing run whose input is empty after filtering and that always saves its result. The dense path's `save()` writes the config without raising, while the keyword and `content: True` paths raised an `AttributeError` from inside `save()` with no hint about the empty input. This change lets the empty keyword index save and load back with `count()` returning `0`. Whether that case is worth carrying is your call.

davidmezzetti:
Thanks. Seems like it's worth handling properly.

davidmezzetti:
LGTM - Thank you!

PR #1226, opened 2026-09-04
fix(workflow): add request timeout to ServiceTask

### LINKED ISSUE #1225: [BUG] ServiceTask requests have no timeout and can hang a workflow indefinitely
**Describe the bug**
`ServiceTask` issues HTTP requests with no timeout. The `requests` library has no default timeout, so a workflow that points a `ServiceTask` at a slow or unresponsive endpoint blocks forever with no way to recover. `ServiceTask.register(...)` also exposes no timeout parameter, so there is no way to bound it from configuration.

**To reproduce**
Define a workflow with a `service` task whose URL never responds (or responds very slowly):

```yaml
workflow:
  slow:
    tasks:
      - task: service
        url: http://<unresponsive-host>/endpoint
        method: get
```

Running the workflow hangs indefinitely.

**Location**
`src/python/txtai/workflow/task/service.py`, in `request()`:

```python
if self.method and self.method.lower() == "get":
    response = requests.get(self.url, params=params)      # no timeout=
else:
    response = requests.post(self.url, json=params)       # no timeout=
```

**Expected behavior**
A slow/unresponsive endpoint should raise a timeout rather than blocking the whole workflow, and the timeout should be configurable.

**Proposed fix / precedent**
There is in-repo precedent for a sane default: the URLRetrieve pipeline (`src/python/txtai/pipeline/data/urlretrieve.py:30`) defaults `timeout=30`. Add a `timeout` parameter (default 30) to `ServiceTask.register(...)`, store it, and forward `timeout=self.timeout` to both the GET and POST calls.

**Environment**
txtai master (@6ebfbbb7)

pacocartones:
Fixes #1225.

### Problem

`ServiceTask` calls `requests.get` / `requests.post` with no `timeout=`. Since `requests` has no default timeout, a workflow whose `ServiceTask` points at a slow or unresponsive endpoint blocks **forever**, and `register(...)` exposed no knob to bound it.

### Fix (minimal, follows existing precedent)

- `ServiceTask.register(...)` gains a `timeout=30` parameter, stored as `self.timeout`.
- Both `requests.get` and `requests.post` now pass `timeout=self.timeout`.

The default and shape mirror the in-repo precedent in the URLRetrieve pipeline (`src/python/txtai/pipeline/data/urlretrieve.py:30`, `timeout=30`). Because `TaskFactory.create` does `TaskFactory.get(task)(**config)`, a `timeout:` key in workflow YAML flows through `Task.__init__` straight to `register(timeout=...)` — so it's configurable per task with a safe default.

### Test

Added `testServiceTimeout` to `test/python/testapi/testapiworkflow.py`: a `/slow` GET branch (`time.sleep(2)`) plus a `slow` workflow with `timeout: 1`, asserting `requests.exceptions.Timeout` is raised.

Validated the real `ServiceTask` class against a live local HTTP server that sleeps 2s with client `timeout=1`:
- **Before:** `register() got an unexpected keyword argument 'timeout'` — no way to bound the call.
- **After:** `requests.exceptions.Timeout` raised after ~1.00s.

`black --check` and `pylint` are clean (10.00/10 on both touched files).

> Note: I couldn't run the full `testapiworkflow` suite locally — its `setUpClass` builds the whole API (sentence-transformers + a labels model, multi-GB downloads) which isn't available in my environment. The added test is syntactically valid and correctly wired (config→`register` path traced); its green run should happen in CI. The standalone real-class reproduction above exercises the actual code path.

### DIFF
--- a/src/python/txtai/workflow/task/service.py
+++ b/src/python/txtai/workflow/task/service.py
@@ -19,7 +19,7 @@ class ServiceTask(Task):
     Task to runs requests against remote service urls.
     """
 
-    def register(self, url=None, method=None, params=None, batch=True, extract=None):
+    def register(self, url=None, method=None, params=None, batch=True, extract=None, timeout=30):
         """
         Adds service parameters to task. Checks if required dependencies are installed.
 
@@ -29,6 +29,7 @@ class ServiceTask(Task):
             params: default query parameters
             batch: if True, all elements are passed in a single batch request, otherwise a service call is executed per element
             extract: list of sections to extract from response
+            timeout: request timeout in seconds, defaults to 30
         """
 
         if not XML_TO_DICT:
@@ -42,6 +43,9 @@ class ServiceTask(Task):
 
         # If True, all elements are passed in a single batch request, otherwise a service call is executed per element
         self.batch = batch
+
+        # Request timeout in seconds. Prevents a slow or unresponsive endpoint from blocking the workflow indefinitely.
+        self.timeout = timeout
 
         # Save sections to extract. Supports both a single string and a hierarchical list of sections.
         self.extract = extract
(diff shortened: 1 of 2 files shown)


davidmezzetti:
LGTM - Thank you!

PR #1227, opened 2026-09-04
Select the top documents in LateEncoder by the document axis

morgan-coded:
`LateEncoder` with `limit` smaller than the number of texts returned the best match followed by arbitrary texts because the partition index was taken from the query axis. This selects the partition index from the document axis, while results for `limit=None` or a limit at least as large as the text count stay unchanged. I reproduced the wrong top-three selection, added a focused test, then ran the full similarity module and pre-commit, all passing.

### DIFF
--- a/src/python/txtai/pipeline/text/lateencoder.py
+++ b/src/python/txtai/pipeline/text/lateencoder.py
@@ -109,7 +109,7 @@ class LateEncoder(Pipeline):
         scores = scores.cpu().numpy()
 
         # Get top n matching indices and scores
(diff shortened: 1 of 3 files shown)


davidmezzetti:
Can the test use the `Similarity` pipeline to be consistent with the other `LateEncoder` tests?

morgan-coded:
Moved the limit coverage onto `Similarity(lateencode=True)` to match the neighboring late-encoder tests. `Similarity` was dropping `limit` on that branch, so it now forwards its keyword arguments to the late encoder the same way the labels branch already does. Nothing else changed.

davidmezzetti:
LGTM - Thank you!

PR #1228, opened 2026-09-04
Fix ANN delete on quantized Torch stores and tiny GGML indexes

morgan-coded:
`Torch` ANN `delete()` raised on `uint8` and `float16` stores because its zero rows were `float32`; a GGML index with one row could not `count()`, `delete()`, or `append()`, and an empty index crashed on load. This makes Torch zero rows use the stored dtype, routes GGML row access through one helper, and rebuilds empty tensors from saved dimensions, while non-empty GGUF files remain byte-identical. I added focused tests for both Torch dtypes and the one-row and empty GGML cases, then ran the dense ANN module and pre-commit with no failures.


davidmezzetti:
LGTM - Thank you!

Issue #1161, opened 2026-07-28
`mcpadapt` not working with latest version of `mcp`


davidmezzetti:
`mcp >= 2.0` has breaking changes that is breaking `mcpadapt`. See https://github.com/grll/mcpadapt/issues/88

AmirF194:
Confirmed at HEAD `cf23682` (clean `python:3.12-slim` container, `pip install '.[agent]'`, resolves `mcp==2.0.0` and `mcpadapt==0.1.20` since `setup.py:55`'s `mcpadapt>=0.1.0` has no upper bound on `mcp`).

The crash is real but it doesn't surface where you'd expect. `txtai.agent.tool.factory` does `import mcpadapt.core` at module level (`factory.py:9`), and `mcpadapt/core.py:20` does `from mcp.client.streamable_http import streamablehttp_client`, which no longer exists under `mcp==2.0.0` (renamed to `streamable_http_client`, matching the fastmcp rename linked from the mcpadapt issue):

```
ImportError: cannot import name 'streamablehttp_client' from 'mcp.client.streamable_http'
```

`txtai/agent/__init__.py:5-10` wraps the whole chain (`.base`, `.factory`, `.tool`) in a bare `try/except ImportError`, and on failure swaps in `placeholder.Agent`, whose `__init__` raises:

```
ImportError: smolagents is not available - install "agent" extra to enable
```

So with the `agent` extra actually installed, `from txtai.agent import Agent; Agent(tools=[...], model=...)` raises exactly that message, which is wrong for this case, the extra is present, `smolagents` imports fine on its own, and the real failure is `mcpadapt` -> `mcp` two imports downstream. I checked this on a bisect of the three `try` imports (`.base`, `.factory`, `.tool` individually): only `.tool` (via `.factory` -> `ToolFactory` -> `mcpadapt.core`) fails, `smolagents` itself imports cleanly under `mcp==2.0.0`.

Fixing this needs either an upstream mcpadapt release (tracked in the linked issue) or a version cap here (`mcpadapt>=0.1.0,<x` or pinning `mcp<2.0.0` in the `agent` extra) as a stopgap, plus possibly narrowing that `except ImportError` so a genuine transitive-dependency break doesn't get relabeled as "extra not installed."

Lori10:
[...]

PR #1233, opened 2026-09-08
fix: surface real import failure in agent placeholder error message (…

1printf:
…#1161)

When agent dependencies fail to import (e.g. mcpadapt incompatible with mcp >= 2.0), the placeholder Agent always raised 'smolagents is not available', which was misleading when the agent extra was already installed.

Preserve the underlying ImportError and include it in the placeholder error message so users see the real cause.

- Save original ImportError in txtai/agent/__init__.py
- Include underlying error in placeholder Agent message
- Add tests for both default and enhanced error messages

### DIFF
--- a/src/python/txtai/agent/__init__.py
+++ b/src/python/txtai/agent/__init__.py
@@ -8,5 +8,10 @@ try:
     from .factory import ProcessFactory
     from .model import PipelineModel
     from .tool import *
-except ImportError:
+except ImportError as e:
     from .placeholder import Agent
+
+    # Preserve the underlying import failure so the placeholder can surface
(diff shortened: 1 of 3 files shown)


davidmezzetti:
Hello - Thanks for the effort on this.

This change doesn't really follow the coding standards for the project so first I suggest you make changes there (i.e. _import_error variable). Second I'd say you should just store the exception then calls raise ImportError(...) from <stored exception>.

davidmezzetti:
Do you plan to work on this?

davidmezzetti:
Going to close this PR. Feel free to resubmit if interested with the changes I mentioned.

PR #1234, opened 2026-09-09
fix(ann): honor quantize: false in the Faiss and SQLite backends

serhiizghama:
The ANN docs say `quantize: false` disables storage quantization ("true sets 8-bit precision, false disables, int sets specified precision"), but the Faiss backend does the opposite. `configure()` maps *any* bool to 8, so an explicit `quantize: false` builds `IDMap,SQ8` instead of `IDMap,Flat` and silently stores 8-bit vectors — the one thing the user asked it not to do. Changed the check to `quantize is True`, which is how the rest of the codebase reads this setting (`numpy.py`, `pgvector.py`, `vectors/base.py` and `transform.py` all exclude bools explicitly).

The same line exists in `sqlite.py`, with the same effect — `sqlite: {quantize: false}` resolved to 8 and turned on `vec_quantize_int8` — so I fixed it there too rather than leave half the backends disagreeing about the same config value.

Tests in `testann/testdense.py`. Both fail on current master (`'IDMap,SQ8' != 'IDMap,Flat'` and `8 != None`) and pass with the change; the rest of `testann.testdense` is unchanged (the errors there are the missing `ann` extra, same before and after).

### DIFF
--- a/src/python/txtai/ann/dense/faiss.py
+++ b/src/python/txtai/ann/dense/faiss.py
@@ -142,7 +142,7 @@ class Faiss(ANN):
 
         # Derive quantization. Prefer backend-specific setting. Fallback to root-level parameter.
         quantize = self.setting("quantize", self.config.get("quantize"))
-        quantize = 8 if isinstance(quantize, bool) else quantize
+        quantize = 8 if quantize is True else quantize
 
         # Get storage setting
         storage = f"SQ{quantize}" if quantize else "Flat"
--- a/src/python/txtai/ann/dense/faiss.py
+++ b/src/python/txtai/ann/dense/faiss.py
@@ -142,7 +142,7 @@ class Faiss(ANN):
 
         # Derive quantization. Prefer backend-specific setting. Fallback to root-level parameter.
         quantize = self.setting("quantize", self.config.get("quantize"))
-        quantize = 8 if quantize is True else quantize
+        quantize = 8 if isinstance(quantize, bool) and quantize else quantize
 
         # Get storage setting
         storage = f"SQ{quantize}" if quantize else "Flat"
--- a/src/python/txtai/ann/dense/sqlite.py
+++ b/src/python/txtai/ann/dense/sqlite.py
@@ -32,7 +32,7 @@ class SQLite(ANN):
 
         # Quantization setting
         self.quantize = self.setting("quantize")
-        self.quantize = 8 if self.quantize is True else int(self.quantize) if self.quantize else None
+        self.quantize = 8 if isinstance(self.quantize, bool) and self.quantize else int(self.quantize) if self.quantize else None
 
     def load(self, path):
         self.path = path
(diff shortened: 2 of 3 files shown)


src/python/txtai/ann/dense/faiss.py
davidmezzetti: This would be better.

```python
        quantize = 8 if isinstance(quantize, bool)  and quantize else quantize
```

src/python/txtai/ann/dense/sqlite.py
```diff
@@ -32,7 +32,7 @@ def __init__(self, config):
 
         # Quantization setting
         self.quantize = self.setting("quantize")
-        self.quantize = 8 if isinstance(self.quantize, bool) else int(self.quantize) if self.quantize else None
+        self.quantize = 8 if self.quantize is True else int(self.quantize) if self.quantize else None
```
davidmezzetti: Same as other comment:

```python
        self.quantize = 8 if isinstance(self.quantize, bool) and self.quantize else int(self.quantize) if self.quantize else None
```

davidmezzetti:
Thanks for the PR. These settings can come in a variety of ways (str, int, bool), it's best to just check if the quantize flag is a bool AND True. Good catch here.

AmirF194 (commented):
I checked whether the isinstance(x, bool) pattern this PR fixes in faiss.py and sqlite.py exists anywhere else, and it does: src/python/txtai/ann/dense/ggml.py:516 has the same shape.

```python
tensortype = self.quantize
tensortype = "Q8_0" if isinstance(tensortype, bool) else f"Q{int(tensortype)}_0" if isinstance(tensortype, int) else tensortype
tensortype = tensortype.upper() if tensortype else "F32"
```

`isinstance(tensortype, bool)` is true for `False` too, so `quantize: false` on the GGML backend picks `Q8_0` instead of `F32`, same symptom this PR describes for Faiss and SQLite. Reproduced the branch in isolation:

```
$ docker run --rm -v /tmp/repro.py:/repro.py python:3.12-slim python3 /repro.py
quantize=False  -> tensortype='Q8_0' expected='F32'   [BUG]
quantize=True   -> tensortype='Q8_0' expected='Q8_0'  [OK]
quantize=8      -> tensortype='Q8_0' expected='Q8_0'  [OK]
```

(repro.py is the three lines above, called with quantize_setting in place of self.quantize)

Same fix shape as the two backends here: `"Q8_0" if tensortype is True else ...`.

serhiizghama:
Done — switched both to `isinstance(quantize, bool) and quantize`.

@AmirF194's note about `ggml.py` holds too: `tensortype()` has the same shape, so `quantize: false` picks `Q8_0` there instead of the documented `F32` default. It needs a slightly different guard than these two — `bool` is an `int` subclass, so `False` drops straight into the `Q{int(...)}_0` branch — so I left it out of this PR. Happy to send it separately if you want it.

AmirF194:
Already opened #1236 for the ggml.py side, thanks for confirming the shape holds there too.

davidmezzetti:

serhiizghama:
Synced this one with master too — it was hitting the same transformers break.

davidmezzetti:
LGTM - Thank you!

PR #1236, opened 2026-09-09
Fix quantize: false silently enabling Q8_0 quantization in the GGML backend

### LINKED ISSUE #1235: GGML backend: quantize: false silently enables Q8_0 quantization
`quantize: false` on the GGML ANN backend does not disable quantization: it still stores embeddings as 8-bit `Q8_0` tensors instead of full-precision `F32`.

`GGMLTensors.tensortype()` (`src/python/txtai/ann/dense/ggml.py`) derives the GGML tensor type from the `quantize` setting:

```python
tensortype = self.quantize
tensortype = "Q8_0" if isinstance(tensortype, bool) else f"Q{int(tensortype)}_0" if isinstance(tensortype, int) else tensortype
tensortype = tensortype.upper() if tensortype else "F32"
```

`isinstance(x, bool)` is true for both `True` and `False`, so an explicit `quantize: false` takes the same branch as `quantize: true` and produces `"Q8_0"` instead of falling through to `"F32"`.

Reproduced against current master by extracting the three lines above verbatim and calling them with each setting:

```
quantize=False  -> tensortype='Q8_0'   expected='F32'    [BUG]
quantize=True   -> tensortype='Q8_0'   expected='Q8_0'   [OK]
quantize=8      -> tensortype='Q8_0'   expected='Q8_0'   [OK]
quantize=4      -> tensortype='Q4_0'   expected='Q4_0'   [OK]
quantize=None   -> tensortype='F32'    expected='F32'    [OK]
```

This is the same `isinstance(x, bool)` shape that #1234 fixes in the Faiss and SQLite backends; `ggml.py` has an identical unguarded case that PR does not touch.

AmirF194:
Fixes #1235.

`GGMLTensors.tensortype()` used `isinstance(quantize, bool)` to detect the default 8-bit quantization case, but that check is true for both `True` and `False`, so `quantize: false` took the same branch as `quantize: true` and produced `Q8_0` instead of falling through to `F32`.

The fix mirrors the `is True` check #1234 uses for the same shape in the Faiss and SQLite backends, combined with the `isinstance(x, int) and not isinstance(x, bool)` guard already used in `numpy.py`/`pgvector.py`/`faiss.py` to keep integer quantization levels (`quantize: 4`, `quantize: 8`, ...) working:

```python
tensortype = "Q8_0" if tensortype is True else f"Q{int(tensortype)}_0" if isinstance(tensortype, int) and not isinstance(tensortype, bool) else tensortype
```

`tensortype()` has exactly two call sites (`createtensors`, `mergetensors`), both of which only consume the returned GGML type constant to build a tensor, so nothing else depends on the old behavior.

Verification:
- `testGGMLQuantizeDisabled` (new): `quantize=False` now resolves to `F32`, `quantize=True` to `Q8_0`, `quantize=4` to `Q4_0`. Fails on master (`False` resolves to `Q8_0`), passes on this branch.
- Full `test.python.testann.testdense` suite (40 tests, all backends): passes.
- `black` and `pylint` (repo's own pre-commit config): clean.
- I did not test against an existing on-disk GGUF index, but `loadtensors()` reads the tensor type back from the file itself rather than calling `tensortype()`, so loading an index saved before this fix should be unaffected; only new writes with `quantize: false` change.

### DIFF
--- a/src/python/txtai/ann/dense/ggml.py
+++ b/src/python/txtai/ann/dense/ggml.py
@@ -519,7 +519,11 @@ class GGMLTensors:
 
         # Read tensor type
         tensortype = self.quantize
-        tensortype = "Q8_0" if isinstance(tensortype, bool) else f"Q{int(tensortype)}_0" if isinstance(tensortype, int) else tensortype
+        tensortype = (
+            "Q8_0"
+            if tensortype is True
+            else f"Q{int(tensortype)}_0" if isinstance(tensortype, int) and not isinstance(tensortype, bool) else tensortype
+        )
         tensortype = tensortype.upper() if tensortype else "F32"
 
         # Validate tensor type
--- a/src/python/txtai/ann/dense/ggml.py
+++ b/src/python/txtai/ann/dense/ggml.py
@@ -521,7 +521,7 @@ class GGMLTensors:
         tensortype = self.quantize
         tensortype = (
             "Q8_0"
-            if tensortype is True
+            if isinstance(tensortype, bool) and tensortype
             else f"Q{int(tensortype)}_0" if isinstance(tensortype, int) and not isinstance(tensortype, bool) else tensortype
         )
         tensortype = tensortype.upper() if tensortype else "F32"
--- a/test/python/testann/testdense.py
+++ b/test/python/testann/testdense.py
@@ -10,12 +10,14 @@ import unittest
 
 from unittest.mock import patch
 
+import ggml
 import numpy as np
 
 from sqlalchemy.dialects.postgresql import BIT
 from sqlalchemy.ext.compiler import compiles
 
 from txtai.ann import ANNFactory, ANN
+from txtai.ann.dense.ggml import GGMLTensors
 from txtai.serialize import SerializeFactory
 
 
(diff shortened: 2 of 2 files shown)


davidmezzetti:
Thank you for catching this. 

I have the same review comments as #1234

```python
        tensortype = "Q8_0" if isinstance(tensortype, bool) and tensortype else f"Q{int(tensortype)}_0" if isinstance(tensortype, int) else tensortype
```

These variables can be strings and ints.

AmirF194:
Ran that exact line against quantize=False: it resolves to Q0_0, and ggml has no GGML_TYPE_Q0_0, so tensortype() would raise ValueError there instead of returning F32 (bool is still a subclass of int, so False falls into the int branch once the first condition is false).

This PR already uses the is True check from your own #1234 review comment (faiss.py and sqlite.py use the identical shape), which handles False, int, and string quantize values correctly, covered by the new test above.

davidmezzetti:
> Ran that exact line against quantize=False: it resolves to Q0_0, and ggml has no GGML_TYPE_Q0_0, so tensortype() would raise ValueError there instead of returning F32 (bool is still a subclass of int, so False falls into the int branch once the first condition is false).
> 
> This PR already uses the is True check from your own #1234 review comment (faiss.py and sqlite.py use the identical shape), which handles False, int, and string quantize values correctly, covered by the new test above.

Please see the other issue. It has been updated and I believe that change should work. Just makes it a bit more clear.

AmirF194:
Switched to `isinstance(tensortype, bool) and tensortype`, matching the shape you settled on in #1234. Functionally identical to the `is True` check (bool only has two values), just consistent with the other two backends now.

davidmezzetti:

AmirF194:
Synced with master (rebase onto 7965a7f4, picking up the transformers<5.17.0 build pin). 40/40 tests pass in testann.testdense, black clean.

davidmezzetti:
LGTM - Thank you!

AmirF194:
Glad it landed. That back and forth on the isinstance edge case was worth it, the fix ended up matching #1234's shape exactly.

PR #1237, opened 2026-09-10
Fix Cluster combining avg() as an unweighted mean of shard averages

AmirF194:
Cluster.search()/batchsearch() recombine each shard's local query results through `Aggregate`, and its avg() was `lambda x: sum(x) / len(x)`, a plain mean of the per-shard averages. That's only correct when every shard contributed the same row count. A shard averaging 100 over 2 rows and a shard averaging 10 over 8 rows combined to 55.0 instead of the true weighted average, 28.0, no error raised. `testSQL` had the same wrong arithmetic baked into its expected value, (6.3+6.7)/2.

avg() now weights by the matching count(*) column, which every avg() query I found in this codebase also selects, and raises SQLError when no count(*) is present to weight by, instead of returning a silently wrong number.

Updated testSQL's expected average to the weighted value and added testAggregateAvg, covering the weighted combine, the single-shard passthrough, and the missing-count(*) SQLError. `python -m unittest testdatabase.testsql testapi.testcluster`: 29 passed, the new unit test fails on main. black and pylint (10.00/10) clean on both changed files.

### DIFF
--- a/src/python/txtai/database/sql/aggregate.py
+++ b/src/python/txtai/database/sql/aggregate.py
@@ -6,6 +6,7 @@ import itertools
 import operator
 
 from .base import SQL
+from .error import SQLError
 
 
 class Aggregate(SQL):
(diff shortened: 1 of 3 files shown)


davidmezzetti:

AmirF194:
Rebased onto master, tests still green (29/29 in testsql/testcluster).

davidmezzetti:
This build error is just a warning about too many public methods. If you add this to testsql it will fix it.

```python
# pylint: disable=R0904
class TestSQL(unittest.TestCase):
```

AmirF194:
Pushed, using your exact snippet. Local pylint run confirms the R0904 is gone.

davidmezzetti:
LGTM - Thank you!

Issue #1078, opened 2026-04-20
Performance Improvement: Use JOIN when JSON fields being used


davidmezzetti:
This change will update the core database query which drives all content-enabled Embeddings queries. When a JSON field filter is enabled, a `JOIN` should be used vs a `LEFT JOIN`. A `LEFT JOIN` requires a full table scan of the `sections` table.

PR #1238, opened 2026-09-10
fix(database): don't drop rows when sorting or grouping by a JSON field

serhiizghama:
`order by` on a JSON field drops rows instead of sorting them.

The JOIN promotion from 8e56946 (#1078) fires on `where`, `groupby` and `orderby`, but only `where` is safe. Plain text passed to `index()` never gets a `documents` row — `loaddocument` writes one only for dicts carrying extra fields — so an inner JOIN removes those sections from the result set entirely.

On a mixed index of one string and two documents with a `length` field:

```
select id from txtai                   ->  0, 1, 2
select id from txtai order by length   ->  1, 2
```

Sorting changed which rows came back. `group by` does the same thing from the other side: the null group vanishes, so aggregates quietly stop counting the text-only rows.

A `where` clause on a JSON field is genuinely different — comparing against NULL already excludes those rows, so the JOIN is a free index win there, and that's how #1078 described it ("when a JSON field filter is enabled"). Left that path alone; the new test asserts it still filters.

Test runs on both the SQLite and DuckDB backends and fails on master with `Lists differ: ['1'] != ['0', '1']`.

### DIFF
--- a/src/python/txtai/database/rdbms.py
+++ b/src/python/txtai/database/rdbms.py
@@ -186,8 +186,9 @@ class RDBMS(Database):
         if indexids:
             select = f"{self.resolve('indexid')}, {self.resolve('score')}"
 
-        # Use JOIN when documents table is used to utilize indexes, default to LEFT JOIN
-        join = "JOIN" if any(x and self.jsonprefix() in x for x in [where, groupby, orderby]) else "LEFT JOIN"
+        # Use JOIN when a JSON field is filtered, since that filter already excludes rows without a
+        # documents row. GROUP BY/ORDER BY don't filter, so they keep the LEFT JOIN.
+        join = "JOIN" if where and self.jsonprefix() in where else "LEFT JOIN"
 
         # Build query text
         query = Statement.TABLE_CLAUSE % (select, join)
--- a/test/python/testdatabase/testrdbms.py
+++ b/test/python/testdatabase/testrdbms.py
@@ -654,6 +654,24 @@ class Common:
 
             self.assertEqual(result["text"], self.data[4])
 
+        def testSortMixed(self):
+            """
+            Test sorting and grouping by a JSON field with mixed text and document data
+            """
+
+            # Mix plain text, which has no documents table row, with a document that has a JSON field
+            embeddings = Embeddings({"keyword": True, "content": self.backend})
+            embeddings.index([(0, self.data[0], None), (1, {"text": self.data[1], "length": 5}, None)])
+
+            # Sorting must not change which rows come back
+            self.assertEqual(sorted(x["id"] for x in embeddings.search("select id from txtai order by length")), ["0", "1"])
+
+            # Grouping keeps a null group for rows missing the field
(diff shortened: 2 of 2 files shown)


davidmezzetti:

serhiizghama:
Done — merged latest master in.

davidmezzetti:
Looking more at this change, the reason LEFT JOIN was added was to utilize indexes. While dropping rows isn't good does this now force full table scans for grouping and sorting?

src/python/txtai/database/rdbms.py
```diff
@@ -186,8 +186,9 @@ def query(self, query, limit, parameters, indexids):
         if indexids:
             select = f"{self.resolve('indexid')}, {self.resolve('score')}"
 
-        # Use JOIN when documents table is used to utilize indexes, default to LEFT JOIN
-        join = "JOIN" if any(x and self.jsonprefix() in x for x in [where, groupby, orderby]) else "LEFT JOIN"
+        # Use JOIN when a JSON field is filtered, since that filter already excludes rows without a
+        # documents row. GROUP BY/ORDER BY don't filter, so they keep the LEFT JOIN.
+        join = "JOIN" if where and self.jsonprefix() in where else "LEFT JOIN"
```
AmirF194, line 191: I ran your full-table-scan question against a scratch SQLite schema matching Statement.CREATE_SECTIONS/CREATE_DOCUMENTS/TABLE_CLAUSE, in a clean python:3.12-slim container, comparing EXPLAIN QUERY PLAN on the old JOIN clause against this PR's LEFT JOIN clause for an ORDER BY on a JSON field.

With an expression index configured on the field (index: true in columns), master drives the sort from documents and uses that index directly:

```
master: JOIN documents, ORDER BY json_extract(data, '$.length')
  SCAN d USING INDEX expression_length
  SEARCH s USING COVERING INDEX section_id (id=?)
```

This PR's LEFT JOIN can't do that. It drives from sections instead and falls back to a temp b-tree sort even with the index present:

```
PR: LEFT JOIN documents, ORDER BY json_extract(data, '$.length')
  SCAN s USING COVERING INDEX section_id
  SEARCH d USING INDEX sqlite_autoindex_documents_1 (id=?) LEFT-JOIN
  USE TEMP B-TREE FOR ORDER BY
```

Without an expression index configured (the default, since columns default to index: false in base.py), both versions already fell back to a temp b-tree sort. The only difference there is which table drives it, documents (master) vs sections (this PR).

Both changes look unavoidable rather than fixable in this PR. The expression index only has entries for rows that made it into documents, so it has no position to sort the section-only rows into once they're included, whichever join keyword is used. And driving from documents is exactly the bug: some sections have no documents row, so the correct query has to visit all of sections to place them. I don't see a way to keep the old plan and also return the rows it was dropping.

I did not check DuckDB's planner, only SQLite.

davidmezzetti: Yah it's unfortunate but I'll have to pick accuracy over speed.

AmirF194: Makes sense, thanks for confirming the tradeoff.

davidmezzetti:
LGTM - Thank you!

PR #1241, opened 2026-09-14
Save IVFSparse indexes after deleting with NumPy ids

morgan-coded:
`IVFSparse.save()` raised a msgpack `TypeError` after a direct ANN caller deleted with a NumPy id array, even though `count()` was correct. This stores deleted ids as Python `int` values, while files written for existing Python-int deletes stay byte-identical. I reproduced the save failure, added a regression that deletes with a NumPy id array, saves, reloads, and checks the count, and ran the sparse ANN module with all 8 tests passing.


davidmezzetti:
LGTM - Thank you!

PR #1242, opened 2026-09-14
Set the HNSW delete counter when loading an index

morgan-coded:
Reopening an HNSW index through the standalone ANN API using a configuration without a delete counter made `count()`, `search()`, and `delete()` raise `KeyError: 'deletes'`. This initializes the missing counter in `load()`, while preserving a counter already supplied in the configuration. I reproduced the missing-counter failure, added a save/load regression, and ran the dense ANN module; the regression passed, and the run of 43 tests had no failures and three skips. This does not recover a lost counter for files with saved deletions; their counts can still be high and searches can still fail above the live row count.


davidmezzetti:
LGTM - Thank you!

PR #1243, opened 2026-09-16
Keep community nodes that match none of the topic terms

linhongyu510:
## What's wrong

`Topics.score()` (`src/python/txtai/graph/topics.py`) replaced the community list
with the result of a term search:

```python
# Sort community by score descending
community = [uid for uid, _ in scoring.search(terms, len(community))]
```

The intent is re-ordering, but `search` is not order-preserving over the input —
it is a filter. `Terms.topn()` ends with:

```python
return [(self.ids[x], float(scores[x])) for x in matches[:limit] if scores[x] > 0]
```

So nodes whose text matches none of the generated topic terms score zero and are
never returned. Passing `len(community)` as the limit prevents truncation but not
this filtering, and the assignment silently shrinks the community.

## Impact

Dropped nodes vanish from topic modeling completely. `Graph.addtopics()` assigns
node attributes by iterating the topic lists:

```python
for x, topic in enumerate(self.topics):
    for r, node in enumerate(self.topics[topic]):
        self.addattribute(node, "topic", topic)
        self.addattribute(node, "topicrank", r)
```

A node that was filtered out therefore has no `topic`, no `topicrank` and no
`category`, and appears under no topic — while still being present in the graph.
Anything reading topics (the `topic`/`topicrank`/`category` expressions in
`Embeddings`, topic-based grouping, RAG over topics) simply cannot see it.

## The existing test recorded the bug

`testCommunity` runs both community algorithms over the same 6-document graph:

```python
graph.config = {"topics": {"algorithm": "greedy"}}
graph.addtopics()
self.assertEqual(sum((len(graph.topics[x]) for x in graph.topics)), 6)

graph.config = {"topics": {"algorithm": "lpa"}}
graph.addtopics()
self.assertEqual(sum((len(graph.topics[x]) for x in graph.topics)), 4)   # <-
```

Six nodes go in, `4` was asserted as the expected number of topic members. Under
`greedy` the count was already 6 because there every node happens to match one of
its community's terms; under `lpa` two nodes were being lost.

Confirmed which ones by running the same setup on an unmodified checkout:

```
graph nodes  : 6
topic members: 4
LOST nodes   : [4, 5]
```

Nodes 4 and 5 are *"Maine man wins $1M from $25 lottery ticket"* and *"Make huge
profits without work, earn up to $100,000 a day"* — `lpa` groups them together,
and neither matches the topic terms generated for that community.

Reproduced directly against `score()` too, which is what `__call__` invokes per
community:

```
community in : ['n1', 'n2', 'n3', 'orphan']
terms        : ['data', 'learning', 'models', 'machine']
community out: ['n1', 'n3', 'n2']
DROPPED      : ['orphan']
```

## The fix

Keep the search result as the ranking, then append whatever it filtered out:

```python
ranked = [uid for uid, _ in scoring.search(terms, len(community))]
unranked = sorted(set(community) - set(ranked), key=lambda x: centrality[x], reverse=True)
community = ranked + unranked
```

Scored nodes keep their existing order and still rank ahead of unmatched ones, so
`topicrank` semantics are unchanged for every node that was already present.
Centrality is the tie-break because that is the ordering `score()` already falls
back to in its own `else` branch when a community has no text at all — no new
concept is introduced.

## Tests

- Updated the `lpa` assertion in `testCommunity` from `4` to `6`. This is the
  behaviour change: all six nodes now land in a topic.
- Added `testCommunityUnmatchedNodes`, which calls `score()` directly with a
  four-node community where one node shares no vocabulary. It asserts no node is
  dropped, and that the unmatched node ranks last — so the fix cannot pass by
  discarding the ranking instead.

The new test uses `GraphFactory.create({})` and needs no model download.

## Verification

```
testCommunityUnmatchedNodes + testCommunity        2 passed
test/python/testgraph.py                          25 passed, 1 failed
black 24.10.0 --check --line-length 150           2 files unchanged
pylint 3.3.1 (repo's pre-commit flags + .pylintrc) 10.00/10 on both files
```

Lint versions are the ones pinned in `.pre-commit-config.yaml`.

The one failure, `testDatabase`, needs the RDBMS dependency from the `graph`
extra, which is not installed here; it fails identically on an unmodified
checkout and does not touch the topics path.

Reverse-verified: with the source change reverted and both tests kept,
`testCommunityUnmatchedNodes` fails and `testCommunity` fails with `4 != 6`;
with the change applied, both pass.

## AI disclosure

This change was prepared with AI assistance. The defect was found by auditing
`Topics.score()` against `Terms.topn()`'s filtering contract, then confirmed by
executing the real classes and by tracing which nodes the existing test was
silently losing; the line references, lost-node identities, lint results and test
counts above were all verified by running the code.


davidmezzetti:
Hello. Thank you for this.

Would you please explain what problem you were facing that led to this PR?

linhongyu510:
@davidmezzetti Thanks for looking. The short version: **`Topics.score()` can silently drop nodes from a community, and those nodes end up with no `topic` / `topicrank` attribute at all.**

I hit it while relying on `topic` being populated for every indexed node — some rows came back with `None`, and the count of nodes covered by topics was lower than the graph's node count.

**The mechanism.** `topics.py` ranks a community by re-searching it with the topic terms:

```python
community = [uid for uid, _ in scoring.search(terms, len(community))]
```

`scoring.search` only returns nodes with a **positive** score. A node that matches none of its community's top-n terms scores zero and is simply absent from the result — and since the result is assigned straight back over `community`, that node is gone. It is a filter wearing the shape of a sort.

**Reproduced directly against `score()`** — four nodes, three sharing vocabulary and one that shares none:

```
community in : ['n1', 'n2', 'n3', 'unmatched']
members  out : ['n1', 'n3', 'n2']
DROPPED      : ['unmatched']
  unmatched: topic=None topicrank=None
```

**This is not only reachable with contrived input — your own test fixture hits it.** That is what the `4` → `6` change in `testCommunity` is. On current `master`, with the six-document fixture already in `testgraph.py`:

```
AssertionError: 4 != 6
```

Two of the six nodes are not in any topic under `lpa`. The old assertion of `4` was encoding the loss as expected behaviour. With the fix, both `greedy` and `lpa` cover all 6, and I confirmed that 6 is the whole graph (`graph.count() == 6`, no node left without a topic) rather than just a bigger number.

**The fix** keeps the ranking, then appends the unscored nodes ordered by centrality:

```python
ranked = [uid for uid, _ in scoring.search(terms, len(community))]
unranked = sorted(set(community) - set(ranked), key=lambda x: centrality[x], reverse=True)
community = ranked + unranked
```

Term-relevant nodes keep their existing order and still rank ahead, so `topicrank` semantics are unchanged for every node that was previously present — the only behavioural change is that nodes which used to vanish now land at the end.

Verification: `test/python/testgraph.py` passes with the fix; with `topics.py` reverted to `master` and the tests kept, `testCommunity` fails with `4 != 6` and `testCommunityUnmatchedNodes` fails with `['n1','n2','n3'] != ['n1','n2','n3','unmatched']`.

Happy to adjust the tie-break for unmatched nodes if you would rather they sort some other way, or to split the `testCommunity` assertion change into its own commit to make the behavioural delta easier to review separately.

davidmezzetti:
Thank you for this. Can you make this comment more succinct?

```python
            # Sort community by score descending. Term search only returns nodes with a
            # positive score, so nodes matching none of the topic terms come back empty.
            # Keep them, ranked last by centrality, otherwise they are dropped from the
            # topic and lose their topic/topicrank attributes entirely.
```

Maybe something like this:

```python
            # Sort community by score descending. Ensure both scored and unscored nodes
            # are kept in community.
```

davidmezzetti:
LGTM - Thank you! I'll make the comment change post merge.

PR #1245, opened 2026-09-18
fix(api): forward zero weights and url-encode parameters in Cluster search

serhiizghama:
`Cluster.search` and `Cluster.batchsearch` only forward `weights` when it's truthy, so `weights=0` (sparse-only on a hybrid index — `Search` treats `0` as "skip dense", the fusion methods check `weights[v] > 0`) never reaches the shards and they fall back to the default 0.5. A single-node `GET /search?weights=0` honors it; the same request against a cluster node quietly searches 50/50.

While there, the GET url in `Cluster.search` was concatenated by hand with only `query` quoted, so the JSON in `parameters` went out raw: a bind value with `&` cuts the JSON off and the shard fails on `json.loads`, `#` drops everything after it (including `graph`), `+` arrives as a space and `%25` as `%`. Building the query string with `urllib.parse.urlencode` fixes that and keeps the wire format the same for the values that already worked. Both changes are covered by tests in `testcluster.py` that look at what the shard actually received; both fail on master.


davidmezzetti:
Thanks for the PR. Could you explain why you added the two lines appending the requests to the tests:

```python
        self.server.requests.append(self.path)
```

```python
        length = int(self.headers.get("content-length", 0))
        self.server.requests.append((self.path, json.loads(self.rfile.read(length)) if length else None))
```

serhiizghama:
Sure. The mock shards return canned responses no matter what comes in, so both bugs are invisible on the response side — a search with `weights=0` or with `&` inside `parameters` still gets the fixture rows back, and a test asserting on the result passes on master too. Recording what each shard actually received is the only way to check that `weights=0.0` made it into the request and that `parameters` arrived intact. The two new tests parse `httpd.requests[-1]` and assert on that; without the capture they'd have nothing to assert on.

I kept it minimal (a list on the server object, appended in `do_GET`/`do_POST`) so the existing handler branches stay as they are. If you'd rather the handler branched on the path and returned distinguishable rows instead, like the `/search?query=select` case does, I can switch to that — it's just harder to make robust for the encoded `parameters` payload.

davidmezzetti:
Got it, sounds good!

davidmezzetti:
LGTM - Thank you!

PR #1248, opened 2026-09-19
Keep deleted keyword results excluded after common-term scoring

### LINKED ISSUE #1247: Mixed rare/common keyword queries can revive deleted results and raise KeyError
### Problem

On master (`19ff237c3758ba7aab9a01fdd576271b4a647fd7`, txtai 9.14.0), a keyword query mixing an uncommon term and a common term can return a deleted document. With `content=True`, resolving that deleted result raises `KeyError` instead.

### Reproduction

No model download is needed:

```python
from txtai.scoring import ScoringFactory

for content in (False, True):
    scoring = ScoringFactory.create({"method": "bm25", "terms": True, "content": content})
    scoring.index([(i, "rare common" if i == 0 else "common", None) for i in range(20)])
    scoring.delete([1])
    try:
        print(scoring.search("rare common", 20))
    finally:
        scoring.close()
```

With `content=False`, the results still contain id 1 with a positive score. With `content=True`, the call raises `KeyError: 1`. Expected: 19 live results, excluding id 1, in both cases.

### Cause and proposed fix

`Terms.candidates()` clears deleted scores before candidate selection. When an uncommon query term was scored, `Terms.topn()` then merges common-term scores into those candidates, which can give deleted candidates positive scores again. This path does not clear them again before sorting and returning results. Queries containing only common terms do clear the scores again through the second `candidates()` call.

I plan to clear deleted scores after the common-term merge, before final ranking, and add regression coverage for mixed/common-only/uncommon-only queries, both content modes, and save/load. This is distinct from #1202 and #1219, which fixed deletion bookkeeping; this reproduction uses one deletion of a unique id.

Reproduced on Windows with Python 3.12.14 and NumPy 2.5.3. Investigation and the planned patch are AI-assisted; test results will be reported explicitly in the PR.

LimbC-C:
## Problem and fix

A keyword query containing both uncommon and common terms can give deleted candidates positive scores again when `Terms.topn()` merges the common-term scores. For example, after deleting one of 20 documents matching `common`, searching `rare common` with a limit of 20 still returns that deleted id. With `content=True`, resolving it raises `KeyError`.

Clear deleted scores again before final sorting, so deleted documents cannot occupy result slots or reach content resolution. This preserves candidate selection and the scoring of live documents.

Fixes #1247.

## Validation

- The new regression test fails on the original implementation in all six method/content combinations: three assertion failures and three `KeyError`s.
- `python -m unittest testscoring.testkeyword -v`: **14 tests passed** on Windows / Python 3.12.14 / NumPy 2.5.3.
- Coverage includes BM25, TF-IDF and SIF; content on/off; uncommon-only, common-only and mixed queries; limits 3 and 20; cached weights; and save/load.
- The same 14 tests pass under coverage, with **100% statement coverage of `txtai.scoring.terms`**.
- Black 24.10.0 and Pylint 3.3.1 passed for both changed files, using the repository configuration and the pre-commit hook's Pylint options. `git diff --check` passed.

Local validation used the minimal installation plus the keyword tests' dependencies. The full model/integration suite and `make coverage` were not run locally. Coverage was collected with a source-directory target; an initial module-name target triggered an import-time NumPy fallback, resolved by switching to the directory target.

AI assistance was used for investigation, implementation and validation.


davidmezzetti:
Thank you for the PR. Does it make sense to clear the deletes at the end of `self.merge` instead?

LimbC-C:
Yes, that is a better place to keep the invariant. Moved the clearing to the end of `merge()` in 9f9b32e, after all common-term scores have been added. The existing regression covers uncommon-only, common-only and mixed queries, both content modes, all three scoring methods, and save/load.

Re-ran all 14 keyword tests successfully; `Terms` statement coverage remains 100%. Black and Pylint checks also pass. Thank you for the suggestion.

davidmezzetti:
LGTM - Thank you!

PR #1249, opened 2026-09-19
Keep SQLite ANN data when saving a modified index to a new path

Irahan2:
Saving a loaded SQLite index to a different path after it was modified silently produced an empty index. With the default float storage, the rows copied into the new database were never committed, so they were rolled back when the connection was closed. With quantized storage (`quantize: 1` / `quantize: 8`), the copy failed instead: `iterdump` writes the vectors as untyped blobs, which sqlite-vec reads as float32 (`invalid float32 vector BLOB length`).

This is the common "load, upsert, save somewhere else" flow:

```python
embeddings = Embeddings()
embeddings.load("index-v1")
embeddings.upsert(documents)     # count() == 100
embeddings.save("index-v2")

embeddings = Embeddings()
embeddings.load("index-v2")      # count() == 0, search() == []
```

This change copies the rows with a `SELECT` / `INSERT` that wraps quantized vectors in their sqlite-vec type (`vec_bit` / `vec_int8`) and commits the new connection. As before, the original database is left untouched.

`testSQLiteCustom` already saves to a new path, but it only checks `count()` on the in-memory model, so it didn't catch this. The new `testSQLiteSaveNewPath` reloads the saved copy for float, bit and int8 storage. It fails on master and passes with this change. I also checked it with the minimum supported sqlite-vec (0.1.1).

AI disclosure: I used Claude Code (Claude Opus 5) to find this bug (by comparing the ANN backends against a brute-force reference through index, delete, append, save and load) and to help write the fix and the test. I reviewed the changes and ran the tests myself.


Irahan2:
Hello @davidmezzetti  , if possibe could you check my PR.

davidmezzetti:
Thanks for the PR. I'll run the build now.

Irahan2:
> Thanks for the PR. I'll run the build now.

Thanks for running it! The Windows job looks like it hit the 60-minute timeout rather than a test failure. All test groups up to testworkflow passed (including the new testSQLiteSaveNewPath), and it was cancelled mid-way through testworkflow. Happy to change anything if needed. 
@davidmezzetti

davidmezzetti:
LGTM - Thank you!

